PR #8, opened 2017-04-04
API-2464 Refactored ApiClient and moved exception handling to API classes

HaloFour:
Cleaned up the implementation of `invoke` in `ApiClient` quite a bit and moved the onus of the exception handling into the API classes.

### DIFF
--- a/resources/sdk/purecloudjava/templates/ApiClient.mustache
+++ b/resources/sdk/purecloudjava/templates/ApiClient.mustache
@@ -559,130 +559,13 @@ public class ApiClient implements AutoCloseable {
         }
     }
 
-    /**
-     * Invoke API by sending HTTP request with the given options.
-     *
-     * @param path         The sub-path of the HTTP URL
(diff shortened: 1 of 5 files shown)


HaloFour:
The big change is that I've eliminated the overloads of `invoke` from `ApiClient`.  Instead there are only two methods: `invoke` and `invokeAsync`.  `ApiClient` no longer worries about the exception munging and instead puts that onus on the API classes directly.  This should correctly handle the following logic:

1. If `shouldThrowException` is `true` then no exceptions are ever thrown.  For methods that return a value, `null` would be returned.  For methods that return an `ApiResponse<T>` an instance wrapping the exception would be returned.
1. If calling a `WithHttpInfo` method or an overload accepting an `ApiRequest<T>` then `ApiException` exceptions are never thrown.  Instead they are converted into `ApiResponse<T>` instances that wrap that exception.
1. In all other cases the `ApiException`, `IOException` or unchecked exceptions would be thrown.

The same logic should apply to both sync and async versions of the API classes.

PR #28, opened 2019-07-01
Adding PureCloudRegionHosts to dotnet

Dschinds:


resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -275,6 +275,29 @@ namespace {{packageName}}.Client
                 return Convert.ToString (obj);
         }
 
+         /// <summary>
+        /// Creates a restclient with a base path string input
+        /// <returns>Returns a rest client</returns>
+        public RestClient withBasePath(String basePath){
```
PrinceNebulon: Name this method `setBasePath`. The "with" naming convention is used for builders, which this is not.

resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -275,6 +275,29 @@ namespace {{packageName}}.Client
...
+            if (String.IsNullOrEmpty(basePath))
+                throw new ArgumentException("basePath cannot be empty");
+
+                 RestClient = new RestClient(basePath);
```
PrinceNebulon: Instead of creating a new RestClient instance, just set the value: `RestClient.BaseUrl = new Uri(basePath);`

resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -275,6 +275,29 @@ namespace {{packageName}}.Client
...
+        }
+        /// <summary>
+        /// Creates a restclient with a PureCloudRegionHost string input
+        /// <returns>Returns a rest client</returns>
+        public RestClient withBasePath(PureCloudRegionHosts region){
+          String basePath = region.GetDescription();
+          if (String.IsNullOrEmpty(basePath))
+                throw new ArgumentException("basePath cannot be empty");
+
+                 RestClient = new RestClient(basePath);
```
PrinceNebulon: Instead of duplicating logic, this method should just call `setBasePath(region.GetDescription())`. Error checking and other logic will only be in the string overload above.

PR #30, opened 2019-07-02
Adding PureCloudRegionHosts

Dschinds:


resources/sdk/purecloudpython/extensions/PureCloudRegionHosts.py
```diff
@@ -0,0 +1,18 @@
...
+    us_east_1 = "https://api.mypurecloud.com"
+    eu_west_1 = "https://api.mypurecloud.ie"
+    ap_southeast_2 = "https://api.mypurecloud.com.au"
+    ap_northeast_1 = "https://api.mypurecloud.jp"
+    eu_central_1 = "https://api.mypurecloud.de"
+
+    def __init__(self, apihost):
+        self.apihost = apihost
+
+    def getapihost(self):
```
PrinceNebulon: Python naming standard for methods is to use underscores. This would be `get_api_host`

resources/sdk/purecloudpython/extensions/PureCloudRegionHosts.py
```diff
@@ -0,0 +1,18 @@
+from enum import Enum
+
+
+class PureCloudRegionHosts(Enum):
```
PrinceNebulon: The name of this class is good, but the filename should be `purecloud_region_hosts.py` to be consistent with the generated files.

PrinceNebulon:
Looks good other than the couple comments about naming standards.

PR #32, opened 2019-07-10
Added a helper function for a saml2bearer token request

Dschinds:
Added the request to ApiClient, updated the pom to include the dependencies. Was able to remove PureCloudRegionHosts from the ApiClient since this is in the same package as APIClient, so it didn't need to be imported. Also included an updated readme.

### DIFF
--- a/resources/sdk/purecloudjava/templates/ApiClient.mustache
+++ b/resources/sdk/purecloudjava/templates/ApiClient.mustache
@@ -19,6 +19,31 @@ import java.time.LocalDate;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.Future;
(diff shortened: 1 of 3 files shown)


resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -174,6 +199,53 @@ public class ApiClient implements AutoCloseable {
 
         return response;
     }
+     public AuthResponse authorizeSaml2Bearer( String clientId, String clientSecret, String orgName, String assertion) throws IOException {
```
PrinceNebulon: Look at authorizeClientCredentials. This should only be a few lines of code.

PR #33, opened 2019-07-11
Add proxy support

asdf123101:
Can be set up in configuration as `PureCloudPlatformClientV2.configuration.proxy = proxy_url`


PrinceNebulon:
Update the readme template with a description of the proxy functionality as well as an example of how to set/use it when using the SDK.

PR #39, opened 2019-07-17
Add doc on how to dump response to JSON

asdf123101:


PrinceNebulon:
Is this something that can be set globally from within the SDK? Are there any default serializers or anything like that you can attach to in python?

asdf123101:
> Is this something that can be set globally from within the SDK? Are there any default serializers or anything like that you can attach to in python?

I think dumping response to JSON is out of the scope of SDK. The SDK is responsible for getting response from API and deserializing it to a corresponding class based on swagger definition. How to handle that response class is up to the developer. So having the documentation in place should be good.

PrinceNebulon:
The purpose of the SDK is to minimize the effort required by API consumers to write their applications. Serialization of data is within that scope. What happens if you call [ApiClient. sanitize_for_serialization(...)](https://github.com/MyPureCloud/platform-client-sdk-python/blob/master/build/PureCloudPlatformClientV2/api_client.py#L254) on a type with a datetime? It looks like that method serializes objects to JSON and supports datetime and date.

asdf123101:
That method is critical for making a request so we should definitely ensure it's working correctly. However, the intention of the SDK, as I suggested in the last comment, is to return a response object. How to serialize/transform the response object is up to the developer. This behavior is consistent to a few other SDKs I checked. For example with JAVA SDK, you need `ObjectMapper` if you want to serialize the response; Javascript can convert JSON to object natively so it doesn't have model definition, but you also need to call `JSON.stringify` to get the raw json response. We could add a `.toRawJson()` interface to the generated APIs but 1. it's most likely not necessary 2. even if we want to add it, we should make it available to all the SDKs.

PrinceNebulon:
RJ verified that `sanitize_for_serialization` does serialize models with datetime properties correctly. Let's document using that since that should handle all cases for all contracts. If you want to add a `toJson` method on the models, that would be useful as well. .NET already has a `toJson` method on models. Java has annotations to make the classes serializable using Jackson, and exposes the objectmapper on apiclient.

asdf123101:
I extracted the `sanitize_for_serialization` to a separate file and used it in the `to_json` method in models. I don't want to import `ApiClient` in models and use that method directly because it can have circular dependency issue.  Do we want to remove this method in `api_client`? It will break existing code if people use this method in their code.

PrinceNebulon:
Looks good, thanks!

PR #47, opened 2020-02-11
API-4076 - Add Retry Logic To Java SDK

deerghayu:
API-4076 - Add Retry Logic To Java SDK

### DIFF
--- a/resources/sdk/purecloudjava/templates/ApiClient.mustache
+++ b/resources/sdk/purecloudjava/templates/ApiClient.mustache
@@ -16,6 +16,7 @@ import java.net.URLEncoder;
 import java.net.Proxy;
 import java.text.DateFormat;
 import java.time.LocalDate;
+import java.time.Duration;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.Future;
@@ -612,7 +613,7 @@ public class ApiClient implements AutoCloseable {
     }
 
     private ApiClientConnectorResponse retryOnFailure(ApiClientConnectorRequest connectorRequest, ApiClientConnectorResponse connectorResponse) {
-        if (connectorResponse.getStatusCode() >= 300) {
+        if (isRetryable(connectorResponse.getStatusCode())) {
             connectorResponse = retryInvoke(connectorRequest, connectorResponse);
         }
         return connectorResponse;
@@ -985,44 +986,60 @@ public class ApiClient implements AutoCloseable {
         public void close() throws Exception { }
     }
 
-    private ApiClientConnectorResponse retryInvoke(ApiClientConnectorRequest connectorRequest, ApiClientConnectorResponse connectorResponse) {
-        int retries = 0;
-        long waitTimeInSec = 1L;
-        boolean retry;
-        int statusCode;
-        int MAX_RETRIES = 0;
-        long MAX_WAIT_INTERVAL_IN_SEC = 32L;
+     private ApiClientConnectorResponse retryInvoke(ApiClientConnectorRequest connectorRequest, ApiClientConnectorResponse connectorResponse) {
+         int retriesBeforeBackOff = 0;
+         long waitTimeInMillis;
+         int backOffRetries = 1;
+         long totalWaitTimeInMillis = 0L;
+         int MAX_RETRIES_BEFORE_BACKOFF = 5;
+         long FIVE_MINUTES_IN_MILLIS = 300000L;
+         long MAX_BACKOFF_IN_MILLIS = 27000L;
+         int MAX_BACKOFF_RETRIES = 3;
 
-        try {
-            final int retry_count = properties.getProperty(ApiClientConnectorProperty.MAX_RETRIES, Integer.class, 0);
-            if (retry_count > 0) {
-                MAX_RETRIES = retry_count;
-            } else {
-                return connectorResponse;
-            }
+         try {
+             do {
+                 if (connectorResponse.getHeaders().containsKey("Retry-After") && retriesBeforeBackOff++ < MAX_RETRIES_BEFORE_BACKOFF) {
+                     waitTimeInMillis = Integer.parseInt(connectorResponse.getHeaders().get("Retry-After")) * 1000L;
 
-            do {
-                connectorResponse = connector.invoke(connectorRequest);
-                statusCode = connectorResponse.getStatusCode();
-                retry = statusCode >= 300;
-                if (retry) {
-                    if (statusCode == 429) {
-                        if (connectorResponse.getHeaders().containsKey("Retry-After")) {
-                            waitTimeInSec = Integer.parseInt(connectorResponse.getHeaders().get("Retry-After"));
-                        }
-                    } else {
-                        waitTimeInSec = Math.min(getWaitTimeExp(retries), MAX_WAIT_INTERVAL_IN_SEC);
-                    }
-                    Thread.sleep(waitTimeInSec * 1000L);
-                }
-            } while (retry && (retries++ < MAX_RETRIES));
-        } catch (Exception exception) {
-            return connectorResponse;
-        }
-        return connectorResponse;
+                 } else if (backOffRetries > MAX_BACKOFF_RETRIES) {
+                     waitTimeInMillis = MAX_BACKOFF_IN_MILLIS;
+
+                 } else {
+                     waitTimeInMillis = getWaitTimeExp(backOffRetries);
+
+                     do {
+                         totalWaitTimeInMillis += waitTimeInMillis;
+                         Thread.sleep(waitTimeInMillis);
+                         connectorResponse = connector.invoke(connectorRequest);
+
+                         if (isFiveMinutesOver(totalWaitTimeInMillis, FIVE_MINUTES_IN_MILLIS)) {
+                             waitTimeInMillis = getWaitTimeExp(++backOffRetries);
+                             totalWaitTimeInMillis = 0L;
+                         }
+                     }
+                     while (isRetryable(connectorResponse.getStatusCode()) && backOffRetries <= MAX_BACKOFF_RETRIES);
+                 }
+
+                 if (isRetryable(connectorResponse.getStatusCode())) {
+                     Thread.sleep(waitTimeInMillis);
+                     connectorResponse = connector.invoke(connectorRequest);
+                 }
+             } while (isRetryable(connectorResponse.getStatusCode()));
+
+         } catch (Exception exception) {
+         }
+         return connectorResponse;
     }
 
     private long getWaitTimeExp(int retryCount) {
-        return (long) Math.pow(2, retryCount);
+        return (long) Math.pow(3, retryCount) * 1000L;
+    }
+
+     private boolean isRetryable(int statusCode) {
+        return Arrays.asList(429, 502, 503, 504).contains(statusCode);
+    }
+
+    private boolean isFiveMinutesOver(long totalWaitTimeInMillis, long fiveMinutesInMillis) {
+        return Duration.ofMillis(fiveMinutesInMillis).compareTo(Duration.ofMillis(totalWaitTimeInMillis)) <= 0;
     }
 }
--- a/resources/sdk/purecloudjava/templates/ApiClient.mustache
+++ b/resources/sdk/purecloudjava/templates/ApiClient.mustache
@@ -1026,6 +1026,7 @@ public class ApiClient implements AutoCloseable {
         public long backoffInterval = 300000L;
         public int defaultDelay = 3;
         public int maxRetryTime = 0;
+        public int maxRetriesBeforeBackoff = 5;
 
         public void setBackoffInterval(long backoffInterval) {
             if (backoffInterval < 0) {
@@ -1047,15 +1048,22 @@ public class ApiClient implements AutoCloseable {
             }
             this.maxRetryTime = maxRetryTime;
         }
+
+        public void setMaxRetriesBeforeBackoff(int maxRetriesBeforeBackoff) {
+            if (maxRetriesBeforeBackoff < 0) {
+                throw new IllegalArgumentException("maxRetriesBeforeBackoff can not be a negative integer");
+            }
+            this.maxRetriesBeforeBackoff = maxRetriesBeforeBackoff;
+        }
     }
 
     public static class Retry {
         private long backoffInterval;
         private int defaultDelay;
         private int maxRetryTime;
-        private final static int MAX_RETRIES_BEFORE_BACKOFF = 5;
+        private int maxRetriesBeforeBackoff;
         private int retryCountBeforeBackOff = 0;
-        private long waitTimeInMillis = defaultDelay * 1000L;
+        private long waitTimeInMillis;
         private Stopwatch stopwatch = null;
 
         private final List<Integer> statusCodes = Arrays.asList(429, 502, 503, 504);
@@ -1065,6 +1073,7 @@ public class ApiClient implements AutoCloseable {
             this.backoffInterval = retryConfiguration.backoffInterval;
             this.defaultDelay = retryConfiguration.defaultDelay;
             this.maxRetryTime = retryConfiguration.maxRetryTime;
+            this.maxRetriesBeforeBackoff = retryConfiguration.maxRetriesBeforeBackoff;
             stopwatch = Stopwatch.createStarted();
         }
 
@@ -1077,9 +1086,11 @@ public class ApiClient implements AutoCloseable {
 
                 if (connectorResponse.getHeaders().containsKey("Retry-After")) {
                     waitTimeInMillis = Integer.parseInt(connectorResponse.getHeaders().getOrDefault("Retry-After", String.valueOf(defaultDelay))) * 1000L;
+                } else {
+                    waitTimeInMillis = defaultDelay * 1000L;
                 }
                 //If status code is 429 then wait until retry-after time and retry. OR If status code is retryable then for the first 5 times: wait until retry-after time and retry.
-                if (connectorResponse.getStatusCode() == 429 || retryCountBeforeBackOff++ < MAX_RETRIES_BEFORE_BACKOFF) {
+                if (connectorResponse.getStatusCode() == 429 || retryCountBeforeBackOff++ < maxRetriesBeforeBackoff) {
                     return waitBeforeRetry(waitTimeInMillis);
                 }
 
--- a/resources/sdk/purecloudjava/templates/ApiClient.mustache
+++ b/resources/sdk/purecloudjava/templates/ApiClient.mustache
@@ -1000,21 +1000,23 @@ public class ApiClient implements AutoCloseable {
     }
 
     public static class Retry {
-        final static long FIVE_MINUTES_IN_MILLIS = 300000L;
-        final static int BASE_BACKOFF_TIME_IN_SEC = 3;
-        final static int MAX_RETRIES_BEFORE_BACKOFF = 5;
-        long totalRetryTimeInMillis = 0L;
-        int retryCountBeforeBackOff = 0;
-        long waitTimeInMillis = BASE_BACKOFF_TIME_IN_SEC * 1000L;
-        Stopwatch stopwatch = null;
-        long timeElapsedPerBucketInMillis = 0L;
+        private final static long FIVE_MINUTES_IN_MILLIS = 300000L;
+        private final static int BASE_BACKOFF_TIME_IN_SEC = 3;
+        private final static int MAX_RETRIES_BEFORE_BACKOFF = 5;
+        private long totalRetryTimeInMillis = 0L;
+        private int retryCountBeforeBackOff = 0;
+        private long waitTimeInMillis = BASE_BACKOFF_TIME_IN_SEC * 1000L;
+        private Stopwatch stopwatch = null;
+        private long timeElapsedPerBucketInMillis = 0L;
+
+        private final List<Integer> statusCodes = Arrays.asList(429, 502, 503, 504);
 
         public boolean shouldRetry(ApiClientConnectorResponse connectorResponse, long timeElapsedPerRequestInMillis) {
             return totalRetryTimeInMillis < getMaxRetryTimeInMillis() && shouldWait(connectorResponse, timeElapsedPerRequestInMillis);
         }
 
         private boolean shouldWait(ApiClientConnectorResponse connectorResponse, long timeElapsedPerRequestInMillis) {
-            if (Arrays.asList(429, 502, 503, 504).contains(connectorResponse.getStatusCode())) {
+            if (statusCodes.contains(connectorResponse.getStatusCode())) {
 
                 if (connectorResponse.getHeaders().containsKey("Retry-After")) {
                     waitTimeInMillis = Integer.parseInt(connectorResponse.getHeaders().getOrDefault("Retry-After", String.valueOf(BASE_BACKOFF_TIME_IN_SEC))) * 1000L;
@@ -1024,32 +1026,30 @@ public class ApiClient implements AutoCloseable {
                     totalRetryTimeInMillis += (waitTimeInMillis + timeElapsedPerRequestInMillis);
                     waitBeforeRetry(waitTimeInMillis);
                     return true;
-
-                    //If status code is 50x and if 3 Sec and 9 Sec buckets were already retried then wait for every 27 Sec and retry.
-                } else if (timeElapsedPerBucketInMillis > (FIVE_MINUTES_IN_MILLIS * 2)) {
+                }
+                //If status code is 50x and if 3 Sec and 9 Sec buckets were already retried then wait for every 27 Sec and retry.
+                if (timeElapsedPerBucketInMillis > (FIVE_MINUTES_IN_MILLIS * 2)) {
                     waitTimeInMillis = getWaitTimeExp(3);
                     totalRetryTimeInMillis += (waitTimeInMillis + timeElapsedPerRequestInMillis);
                     waitBeforeRetry(waitTimeInMillis);
                     return true;
-
-                    //If status code is 50x then wait for every 3 Sec and retry until 5 minutes then after wait for every 9 Sec and retry until next 5 minutes.
-                } else {
-                    if (stopwatch == null) {
-                        stopwatch = Stopwatch.createStarted();
-                    }
-
-                    if (timeElapsedPerBucketInMillis < FIVE_MINUTES_IN_MILLIS) {
-                        waitTimeInMillis = getWaitTimeExp(1);
-
-                    } else if (timeElapsedPerBucketInMillis < (FIVE_MINUTES_IN_MILLIS * 2)) {
-                        waitTimeInMillis = getWaitTimeExp(2);
-                    }
-                    waitBeforeRetry(waitTimeInMillis);
-                    timeElapsedPerBucketInMillis = stopwatch.elapsed(TimeUnit.MILLISECONDS) + waitTimeInMillis + timeElapsedPerRequestInMillis;
-                    totalRetryTimeInMillis += (waitTimeInMillis + timeElapsedPerRequestInMillis);
-
-                    return true;
                 }
+                //If status code is 50x then wait for every 3 Sec and retry until 5 minutes then after wait for every 9 Sec and retry until next 5 minutes.
+                if (stopwatch == null) {
+                    stopwatch = Stopwatch.createStarted();
+                }
+
+                if (timeElapsedPerBucketInMillis < FIVE_MINUTES_IN_MILLIS) {
+                    waitTimeInMillis = getWaitTimeExp(1);
+
+                } else if (timeElapsedPerBucketInMillis < (FIVE_MINUTES_IN_MILLIS * 2)) {
+                    waitTimeInMillis = getWaitTimeExp(2);
+                }
+                waitBeforeRetry(waitTimeInMillis);
+                timeElapsedPerBucketInMillis = stopwatch.elapsed(TimeUnit.MILLISECONDS) + waitTimeInMillis + timeElapsedPerRequestInMillis;
+                totalRetryTimeInMillis += (waitTimeInMillis + timeElapsedPerRequestInMillis);
+
+                return true;
             }
             return false;
         }
--- a/resources/sdk/purecloudjava/templates/ApiClient.mustache
+++ b/resources/sdk/purecloudjava/templates/ApiClient.mustache
@@ -1021,7 +1021,7 @@ public class ApiClient implements AutoCloseable {
         return connectorResponse;
     }
 
-    public long getWaitTimeExp(int retryCount) {
+    private long getWaitTimeExp(int retryCount) {
         long waitTime = ((long) Math.pow(2, retryCount) * 1000L);
         return waitTime;
     }
--- a/resources/sdk/purecloudjava/templates/ApiClient.mustache
+++ b/resources/sdk/purecloudjava/templates/ApiClient.mustache
@@ -87,7 +87,13 @@ public class ApiClient implements AutoCloseable {
     private int maxRetryTimeInSec = 0;
     private final long FIVE_MINUTES_IN_MILLIS = 300000L;
     private final int BASE_BACKOFF_TIME_IN_SEC = 3;
-    private final int MAX_BACKOFF_TIME_IN_SEC = 27;
+    private long totalRetryTimeInMillis = 0L;
+    private int bucketCount = 1;
+    private int maxBuckets = 3;
+    private int retryCountBeforeBackOff = 0;
+    private long waitTimeForEachBucketInMillis = 0L;
+    private int maxRetriesBeforeBackOff = 5;
+    private long waitTimeInMillis = BASE_BACKOFF_TIME_IN_SEC * 1000L;
 
     public ApiClient() {
         this(Builder.standard());
@@ -571,10 +577,12 @@ public class ApiClient implements AutoCloseable {
         ApiClientConnectorRequest connectorRequest = prepareConnectorRequest(request, isAuthRequest);
         ApiClientConnectorResponse connectorResponse = null;
         try {
-            connectorResponse = connector.invoke(connectorRequest);
-             if (getMaxRetryTimeInMillis() > 0 && shouldRetry(connectorResponse.getStatusCode())) {
-                 connectorResponse = retryInvoke(connectorRequest, connectorResponse);
-             }
+             connectorResponse = connector.invoke(connectorRequest);
+        if (getMaxRetryTimeInMillis() > 0) {
+            do {
+                connectorResponse = connector.invoke(connectorRequest);
+            } while (totalRetryTimeInMillis < getMaxRetryTimeInMillis() && shouldRetry(connectorResponse));
+        }
             return interpretConnectorResponse(connectorResponse, returnType);
         }
         finally {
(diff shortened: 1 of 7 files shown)


resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +984,48 @@ public class ApiClient implements AutoCloseable {
...
+                statusCode = connectorResponse.getStatusCode();
+
+                if (statusCode >= 300) {
+                    if (connectorResponse.getHeaders().containsKey("Retry-After")) {
+                        waitTime = Integer.parseInt(connectorResponse.getHeaders().get("Retry-After"));
+                    }
+
+                    if (statusCode == 429) {
+                        retry = true;
+                        Thread.sleep(waitTime * 100L);
```
PrinceNebulon: Is this correct? The retry-after header is in seconds and thread.sleep expects miliseconds, which should mean multiplying by 1000, not 100.

deerghayu: Technically it should be 1000 but while testing I was getting two-digit number from V2 all the time so to make it a realistic wait time I made it 100. But now making it 1000 again.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +984,45 @@ public class ApiClient implements AutoCloseable {
...
         public void close() throws Exception { }
     }
+
+    private ApiClientConnectorResponse retryInvoke(ApiClientConnectorRequest connectorRequest, ApiClientConnectorResponse connectorResponse) {
+        int retries = 0;
+        long waitTimeInSec = 1L;
+        boolean retry;
+        int statusCode;
+        int MAX_RETRIES = 0;
+        long MAX_WAIT_INTERVAL_IN_SEC = 32L;
```
deerghayu: Please suggest what should be the appropriate value for MAX_WAIT_INTERVAL_IN_SEC ?

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -585,6 +586,7 @@ public class ApiClient implements AutoCloseable {
                     try {
                         ApiResponse<T> response;
                         try {
+                            connectorResponse = retryOnFailure(connectorRequest, connectorResponse);
```
deerghayu: Adding Async version shortly.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -609,6 +611,13 @@ public class ApiClient implements AutoCloseable {
         return future;
     }
 
+    private ApiClientConnectorResponse retryOnFailure(ApiClientConnectorRequest connectorRequest, ApiClientConnectorResponse connectorResponse) {
+        if (connectorResponse.getStatusCode() >= 300) {
+            connectorResponse = retryInvoke(connectorRequest, connectorResponse);
```
chuckpulfer: 400 level errors are not retry-able (except 429). I would only retry on a 502, 503, or 504 response codes.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +984,45 @@ public class ApiClient implements AutoCloseable {
...
+                connectorResponse = connector.invoke(connectorRequest);
+                statusCode = connectorResponse.getStatusCode();
+                retry = statusCode >= 300;
+                if (retry) {
+                    if (statusCode == 429) {
+                        if (connectorResponse.getHeaders().containsKey("Retry-After")) {
+                            waitTimeInSec = Integer.parseInt(connectorResponse.getHeaders().get("Retry-After"));
+                        }
+                    } else {
+                        waitTimeInSec = Math.min(getWaitTimeExp(retries), MAX_WAIT_INTERVAL_IN_SEC);
```
chuckpulfer: The first few retries should honor the retry-after header if it's populated. Then fallback to a back-off if the failures continue.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -561,6 +561,7 @@ public class ApiClient implements AutoCloseable {
         ApiClientConnectorResponse connectorResponse = null;
         try {
             connectorResponse = connector.invoke(connectorRequest);
+            connectorResponse = retryOnFailure(connectorRequest, connectorResponse);
```
Mikosz91: Wouldn't you want to only enter this method if the request failed with a 429 or 5xx?

resources/sdk/purecloudjava/templates/ApiClient.mustache
Mikosz91: Yeah, I would switch on the status code probably

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +984,45 @@ public class ApiClient implements AutoCloseable {
...
+    private ApiClientConnectorResponse retryInvoke(ApiClientConnectorRequest connectorRequest, ApiClientConnectorResponse connectorResponse) {
+        int retries = 0;
+        long waitTimeInSec = 1L;
+        boolean retry;
+        int statusCode;
+        int MAX_RETRIES = 0;
+        long MAX_WAIT_INTERVAL_IN_SEC = 32L;
+
+        try {
+            final int retry_count = properties.getProperty(ApiClientConnectorProperty.MAX_RETRIES, Integer.class, 0);
```
Mikosz91: You could just assign MAX_RETRIES here and return if it's 0, right?

resources/sdk/purecloudjava/templates/ApiClient.mustache
Mikosz91: If it's a 429, use w/e the value is. If it's a 5xx, how about 60s?

resources/sdk/purecloudjava/templates/ApiClient.mustache
PrinceNebulon: @chuckpulfer The current code should only retry if the request is not a 1xx or 2xx **and** there's a retry-after header. Anything that doesn't have a retry-after header won't be retried. I'd prefer for the SDK to be dumb in this case and not check for specific response codes and let the API tell the SDK when it's ok to retry requests by sending the retry-after header. 429 is being checked specifically since we want different behavior in that case. This allows the SDK to be future proof for any changes made to the API regarding when it sends retry-after headers. Do you want the SDK to ignore retry-after headers if the status code isn't 429, 502, 503, or 504?

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +985,61 @@ public class ApiClient implements AutoCloseable {
...
+                     while (isRetryable(connectorResponse.getStatusCode()) && backOffRetries <= MAX_BACKOFF_RETRIES);
+                 }
+
+                 if (isRetryable(connectorResponse.getStatusCode())) {
+                     Thread.sleep(waitTimeInMillis);
+                     connectorResponse = connector.invoke(connectorRequest);
+                 }
+             } while (isRetryable(connectorResponse.getStatusCode()));
+
+         } catch (Exception exception) {
```
Tom-J-Barrett: Where may the exception be throw?

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +985,61 @@ public class ApiClient implements AutoCloseable {
...
+         long MAX_BACKOFF_IN_MILLIS = 27000L;
+         int MAX_BACKOFF_RETRIES = 3;
+
+         try {
+             do {
+                 if (connectorResponse.getHeaders().containsKey("Retry-After") && retriesBeforeBackOff++ < MAX_RETRIES_BEFORE_BACKOFF) {
+                     waitTimeInMillis = Integer.parseInt(connectorResponse.getHeaders().get("Retry-After")) * 1000L;
+
+                 } else if (backOffRetries > MAX_BACKOFF_RETRIES) {
+                     waitTimeInMillis = MAX_BACKOFF_IN_MILLIS;
```
Mikosz91: A 429 wouldn't need/what this max backoff retries applied. If the request is a 429, grab the retry-after value and invoke. The subsequent response could still be evaluated but the MAX_BACKOFF_RETRIES concept should be removed.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +985,61 @@ public class ApiClient implements AutoCloseable {
         @Override
         public void close() throws Exception { }
     }
+
+     private ApiClientConnectorResponse retryInvoke(ApiClientConnectorRequest connectorRequest, ApiClientConnectorResponse connectorResponse) {
+         int retriesBeforeBackOff = 0;
```
Mikosz91: Is this the retries before a fixed backoff bucket? Consider tweaking the name a little bit for clarity

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +985,61 @@ public class ApiClient implements AutoCloseable {
...
+         long totalWaitTimeInMillis = 0L;
+         int MAX_RETRIES_BEFORE_BACKOFF = 5;
+         long FIVE_MINUTES_IN_MILLIS = 300000L;
+         long MAX_BACKOFF_IN_MILLIS = 27000L;
+         int MAX_BACKOFF_RETRIES = 3;
+
+         try {
+             do {
+                 if (connectorResponse.getHeaders().containsKey("Retry-After") && retriesBeforeBackOff++ < MAX_RETRIES_BEFORE_BACKOFF) {
+                     waitTimeInMillis = Integer.parseInt(connectorResponse.getHeaders().get("Retry-After")) * 1000L;
```
Mikosz91: Will this throw an exception if the retry after isn't set? That'd be possible if Public API is totally down

deerghayu: yes, it may throw an exception during parsing and the original connectorResponse will be returned.

resources/sdk/purecloudjava/templates/ApiClient.mustache
deerghayu: As per my understanding, we don't want to throw an exception from here instead return the original response.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +985,61 @@ public class ApiClient implements AutoCloseable {
...
+
+    private long getWaitTimeExp(int retryCount) {
+        return (long) Math.pow(3, retryCount) * 1000L;
+    }
+
+     private boolean isRetryable(int statusCode) {
+        return Arrays.asList(429, 502, 503, 504).contains(statusCode);
+    }
+
+    private boolean isFiveMinutesOver(long totalWaitTimeInMillis, long fiveMinutesInMillis) {
```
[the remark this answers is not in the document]
deerghayu: I tried to convert all the numbers into Millis to remove confusion. I'll rename the method to align.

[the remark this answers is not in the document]
Tom-J-Barrett: Thanks :)

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -585,6 +593,7 @@ public class ApiClient implements AutoCloseable {
                     try {
                         ApiResponse<T> response;
                         try {
+                           //connectorResponse = retryOnFailure(connectorRequest, connectorResponse);
```
PrinceNebulon: Remove commented out code

PrinceNebulon:
Still need to do:

1. The ApiClient builder needs to have a configuration option to enable/disable or configure the retry feature, and it needs to default to off. 
2. The `README.md` template needs to be updated to document how to enable the retry logic and should explain exactly what behavior to expect when it's enabled.

resources/sdk/purecloudjava/templates/ApiClient.mustache
deerghayu: yes. Going to add Async version to that.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +984,62 @@ public class ApiClient implements AutoCloseable {
...
+                        connectorResponse = waitAndRetry(waitTimeInMillis, connectorRequest);
+
+                         //for every 5 minutes switch the bucket from 3 Sec through 27 Sec and reset the wait time for each bucket.
+                        if (isFiveMinutesOver(waitTimeForEachBucketInMillis)) {
+                            waitTimeInMillis = getWaitTimeExp(++bucketCount);
+                            waitTimeForEachBucketInMillis = 0L;
+                        }
+                    } while (shouldRetry(connectorResponse.getStatusCode()) && bucketCount <= maxBuckets);
+                }
+            } catch (Exception exception) {
```
PrinceNebulon: I'm not sure we should suppress the error here. This will allow a systemic problem with the retry process to go silently unnoticed. I would suggest simply removing the try/catch to let it be thrown out of this method. Under normal retry processing, exceptions shouldn't be thrown and this code will complete gracefully after reaching max retries.

resources/sdk/purecloudjava/templates/README.mustache
```diff
@@ -170,6 +170,19 @@ UserMe me = usersApi.getUsersMe(Collections.singletonList("presence"));
 System.out.println("Hello " + me.getName());
 ```
 
+## How to Configure Retry with Java SDK
```
PrinceNebulon: These docs should be at line 130 at heading level 4 just before `#### Other ApiClient.Builder methods`, since it's documenting a feature of the ApiClient's builder. The heading should be "Setting the max retry time" to fit in with the way the other headings are worded.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -560,7 +577,12 @@ public class ApiClient implements AutoCloseable {
         ApiClientConnectorRequest connectorRequest = prepareConnectorRequest(request, isAuthRequest);
         ApiClientConnectorResponse connectorResponse = null;
         try {
-            connectorResponse = connector.invoke(connectorRequest);
+             connectorResponse = connector.invoke(connectorRequest);
```
PrinceNebulon: This line and the following if block can be removed since the do block will always be executed at least once. This will also fix a bug because the current code will always make the first retry immediately without waiting. 

Additionally, all of the request-specific properties need to be refactored out of ApiClient and into ApiClientConnectorRequest. The request object itself should be used to track those values because Java apps can be multi-threaded. Making concurrent requests would result in props like totalRetryTimeInMillis being added to by multiple requests.

On the topic of totalRetryTimeInMillis, it would be a more accurate approach to instead set a timestamp on the request right before it's made and measure the time since that timestamp to determine which bucket you're in. That provides a few benefits:

* More accurate because it won't exclude the time it takes to execute the request
* Removes the need for the variables tracking bucket stuff. You you can replace most of that logic with a check on the time since the started timestamp to see if duration < 5 mins, if < 10 mins, or > 10.
* Allows you to remove the initial check for the configured max retry time (mentioned above) because the measured running time will always be greater than 0 after the first request.

resources/sdk/purecloudjava/templates/README.mustache
```diff
@@ -128,6 +128,19 @@ Specify the connector in the builder:
...
+#### Setting the max retry time
+
+In default, Java SDK makes a call to API without a Retry option but this feature is configurable.
+The SDK configuration is default to Retry off and it can be configured to enable or disable while creating an `ApiClient` Instance.
+
+Specify the amount of time in seconds that client wants to wait while retrying if client opts to enable Retry feature:
+```{"language":"java"}
+.withMaxRetryTimeInSec(30)
+```
+When Retry feature is enabled and if the `status code` form the API response is `429 or 502 or 503 or 504`,
```
PrinceNebulon: Remove the rest of this explanation and replace with:

```
When the retry feature is enabled, the SDK adheres to the recommended best practices for automatically retrying retryable responses. The best practices are documented in the [Rate Limiting](https://developer.mypurecloud.com/api/rest/rate_limits.html) Developer Center article.
```

deerghayu: done.

resources/sdk/purecloudjava/templates/ApiClient.mustache
deerghayu: Refactored accordingly. 
For now, I've created an inner class to make it threadsafe allowing per request/thread to create a new instance of the inner class. It can be extracted out to a separate class if necessary.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1031,72 @@ public class ApiClient implements AutoCloseable {
...
+                totalRetryTimeInMillis += (waitTimeInMillis + timeElapsedPerRequestInMillis);
+
+                return true;
+            }
+            return false;
+        }
+
+        private void waitBeforeRetry(long waitTimeInMillis) {
+            try {
+                Thread.sleep(waitTimeInMillis);
```
chuckpulfer: Could we log this at info?

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -558,17 +567,24 @@ public class ApiClient implements AutoCloseable {
 
     private <T> ApiResponse<T> getAPIResponse(ApiRequest<?> request, TypeReference<T> returnType, boolean isAuthRequest) throws IOException, ApiException {
         ApiClientConnectorRequest connectorRequest = prepareConnectorRequest(request, isAuthRequest);
+        Retry retry = new Retry();
         ApiClientConnectorResponse connectorResponse = null;
+        long timeElapsedPerRequestInMills = 0L;
         try {
-            connectorResponse = connector.invoke(connectorRequest);
+            do {
+                Stopwatch timeElapsedPerRequest = Stopwatch.createStarted();
```
PrinceNebulon: Don't track the time of individual requests. Create the stopwatch in the Retry constructor and store it in Retry. Then use the elapsedTime method on the stopwatch to check how long it's been since the process started. No updating, no math, no logic except a check against the elapsed time.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -583,8 +599,16 @@ public class ApiClient implements AutoCloseable {
                 @Override
                 public void onCompleted(ApiClientConnectorResponse connectorResponse) {
                     try {
+                        ApiClient.Retry retry = new ApiClient.Retry();
+                        long timeElapsedPerRequestInMills = 0L;
                         ApiResponse<T> response;
                         try {
+                            do {
```
PrinceNebulon: I don't think there should be a loop here. This code will invoke the async API infinitely because the connectorResponse variable that's being checked is never updated here.

resources/sdk/purecloudjava/templates/ApiClient.mustache
deerghayu: By doing this it will also include the time spent for the first 5 retries for 50x before coming to 5 minutes buckets (intermediate 429) while calculating 5 minutes buckets time. If we ignore that we can use one stopwatch across the process.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -762,6 +813,11 @@ public class ApiClient implements AutoCloseable {
             return this;
         }
 
+         public Builder withMaxRetryTimeInSec(int maxRetryTimeInSec) {
```
PrinceNebulon: In preparation for unit testing, it would be a good pattern to expand this concept to accept a RetryConfiguration object instead of a single value. These values can be passed into the Retry object's constructor. The RetryConfiguration object would contain settings for:

* backoffInterval - defaults to 5 minutes, replaces FIVE_MINUTES_IN_MILLIS
* defaultDelay - defaults to 3 seconds. replaces BASE_BACKOFF_TIME_IN_SEC
* maxRetryTime - defaults to 0, replaces existing config option

This will allow the unit tests to validate that the backoff logic works on a much shorter timescale.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1031,72 @@ public class ApiClient implements AutoCloseable {
...
+                    totalRetryTimeInMillis += (waitTimeInMillis + timeElapsedPerRequestInMillis);
+                    waitBeforeRetry(waitTimeInMillis);
+                    return true;
+                }
+                //If status code is 50x and if 3 Sec and 9 Sec buckets were already retried then wait for every 27 Sec and retry.
+                if (timeElapsedPerBucketInMillis > (FIVE_MINUTES_IN_MILLIS * 2)) {
+                    waitTimeInMillis = getWaitTimeExp(3);
+                    totalRetryTimeInMillis += (waitTimeInMillis + timeElapsedPerRequestInMillis);
+                    waitBeforeRetry(waitTimeInMillis);
+                    return true;
```
PrinceNebulon: A lot of the code in this class can be streamlined. For example, these four can be reduced to one by making waitBeforeRetry return true and replacing these lines with: `return waitBeforeRetry(getWaitTimeExp(3))`. Tracking the total time via math is unnecessary as mentioned abvove.

resources/sdk/purecloudjava/templates/ApiClient.mustache
PrinceNebulon: Yes, that's what I'm asking you to do. Measure the **total** time since the first request was made. All you need to do is measure if stopwatch.elapsed() < 5 minutes, < 10 minutes, or else to determine which value to use for the delay. This might even work with a single expression using something like this (pseudocode, won't compile): `double delay = Math.pow(duration, Math.floor((stopwatch.elapsed(seconds) / (60 * 5)) + 1);`.

resources/sdk/purecloudjava/templates/ApiClient.mustache
deerghayu: log added.

PrinceNebulon: I think he meant logging that the SDK will be sleeping for X duration.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1055,79 @@ public class ApiClient implements AutoCloseable {
...
+
+                if (connectorResponse.getHeaders().containsKey("Retry-After")) {
+                    waitTimeInMillis = Integer.parseInt(connectorResponse.getHeaders().getOrDefault("Retry-After", String.valueOf(defaultDelay))) * 1000L;
+                }
+                //If status code is 429 then wait until retry-after time and retry. OR If status code is retryable then for the first 5 times: wait until retry-after time and retry.
+                if (connectorResponse.getStatusCode() == 429 || retryCountBeforeBackOff++ < MAX_RETRIES_BEFORE_BACKOFF) {
+                    return waitBeforeRetry(waitTimeInMillis);
+                }
+                //If status code is 50x and if 3 Sec and 9 Sec buckets were already retried then wait for every 27 Sec and retry.
+                if (stopwatch.elapsed(TimeUnit.MILLISECONDS) > (backoffInterval * 2)) {
```
PrinceNebulon: These three conditions can be replaced with one line: `return waitBeforeRetry(getWaitTimeExp(Math.min(3, Math.floor(stopwatch.elapsed(TimeUnit.MILLISECONDS)/backoffInterval) + 1)))`

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -792,6 +838,40 @@ public class ApiClient implements AutoCloseable {
         }
     }
 
+    public static class RetryConfigBuilder {
```
PrinceNebulon: Having a builder for 3 properties isn't necessary, but if you want to keep it, it should be created using the same pattern as the API client's builder with a private constructor that is only used by the `standard()` static method.

resources/sdk/purecloudjava/templates/ApiClient.mustache
deerghayu: sorry, got it.

resources/sdk/purecloudjava/templates/ApiClient.mustache
deerghayu: perfect. thank you.

resources/sdk/purecloudjava/tests/ApiClientRetryTest.java
```diff
@@ -0,0 +1,337 @@
...
+        };
+
+        ApiClient.RetryConfiguration retryConfiguration = new ApiClient.RetryConfiguration();
+        retryConfiguration.setMaxRetryTime(30);
+
+        ApiClient.Retry retry = new ApiClient.Retry(retryConfiguration);
+        Stopwatch stopwatch = Stopwatch.createStarted();
+        boolean result = retry.shouldRetry(response);
+        Assert.assertEquals(true, result, "Status Code is 429, so it will sleep for 1 Sec as provided in Retry-After header and returns true");
+        Assert.assertTrue(stopwatch.elapsed(TimeUnit.MILLISECONDS) < 1100);
```
Tom-J-Barrett: Should we also assert that the elapsed time is over a certain threshold?

deerghayu: I think we can do both way but asserting to a calculated time period is more precise than asserting to over a certain threshold..

PrinceNebulon: @Tom-J-Barrett yes

@deerghayu per our conversation yesterday, this should be checking for the elapsed time to be within 100ms of the expected runtime.

deerghayu: oh I see I what I was missing. thanks guys

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -644,9 +669,7 @@ public class ApiClient implements AutoCloseable {
     }
 
     public <T> Future<ApiResponse<T>> invokeAsync(ApiRequest<?> request, TypeReference<T> returnType, AsyncApiCallback<ApiResponse<T>> callback) {
-        SettableFuture<ApiResponse<T>> future = SettableFuture.create();
-        getAPIResponseAsync(request, returnType, callback);
-        return future;
+        return getAPIResponseAsync(request, returnType, callback);
```
deerghayu: I refactored this assuming `future` is never set that caused main thread waiting.

resources/sdk/purecloudjava/tests/ApiClientRetryTest.java
```diff
@@ -0,0 +1,445 @@
...
+            builder = builder.withProperty(ApiClientConnectorProperty.CONNECTOR_PROVIDER, connector);
+            if (useenum) {
+                builder = builder.withBasePath(region);
+            } else {
+                builder = builder.withBasePath("https://api." + environment);
+            }
+            builder = builder.withRetryConfiguration(retryConfiguration);
+            apiClient = builder.build();
+
+            ApiResponse<AuthResponse> authResponse = apiClient.authorizeClientCredentials(getClientId(), getClientSecret());
```
PrinceNebulon: The tests are failing here with an NPE.

```
[INFO] Running TestSuite
[TEST] beforeTest
Before test
[SUCCESS] beforeTest
[TEST] setup
[SUCCESS] setup
[TEST] invokeTestWith_429
Not in PureCloudRegionHosts using string value
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
java.lang.NullPointerException
[FAILED] invokeTestWith_429
[ERROR] Tests run: 53, Failures: 18, Errors: 0, Skipped: 35, Time elapsed: 0.904 s <<< FAILURE! - in TestSuite
[ERROR] invokeTestWith_429(com.mypurecloud.sdk.v2.ApiClientRetryTest)  Time elapsed: 0.236 s  <<< FAILURE!
java.lang.AssertionError: null
	at com.mypurecloud.sdk.v2.ApiClientRetryTest.getApiClient(ApiClientRetryTest.java:299)
	at com.mypurecloud.sdk.v2.ApiClientRetryTest.invokeTestWith_429(ApiClientRetryTest.java:53)
```

deerghayu: I always used AccessToken to test against local because with clientId and clientSecret It would run against https://apps.inindca.com.

But adding this as like in the Integration test(SdkTests).

PrinceNebulon: The tests must authenticate using client credentials. The test is configured correctly to do that, but the NPE is coming from your retry code somewhere. use `ex.PrintStackTrace()` to see exactly where. Make sure all of the tests are passing against DCA; none of the tests should rely on the API _actually_ responding with an error; those should be mocked.

deerghayu: NPE occurred due to mocked response on API call during authentication that was used for both Authentication and actual API call.

`Mockito.spy()` fixed the issue with a real call for authentication and with a mocked response for Actual API call.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1005,95 @@ public class ApiClient implements AutoCloseable {
         @Override
         public void close() throws Exception { }
     }
+
+    public static class RetryConfiguration {
+        public long backoffInterval = 300000L;
+        public long defaultDelay = 3000L;
+        public int maxRetryTime = 0;
+        public int maxRetriesBeforeBackoff = 5;
```
Mikosz91: @PrinceNebulon is there any way you can think of to protect these configuration options from being set outside of the tests? I don't see them in this review, but I was thinking we could make these static final constants but add some getters that the tests could mock

Mikosz91: Ah, they just didn't load for me. I see them now

resources/sdk/purecloudjava/tests/ApiClientRetryAsyncTest.java
```diff
@@ -0,0 +1,510 @@
...
+            public void onCompleted(ApiResponse<ApiClientConnectorResponse> response) {
+            }
+
+            @Override
+            public void onFailed(Throwable exception) {
+            }
+        };
+    }
+
+    private CloseableHttpResponse getCloseableHttpResponse(int statusCode, Map<String, String> headerMap) {
```
Mikosz91: Can this be a mock instead of a concrete impl?

deerghayu: Could be but I wanted to use different status codes and headers for different tests so making it generic.
couldn't help myself for implementing all the unwanted methods though :(

deerghayu: I think I can not. once I create a mocked object: 
`CloseableHttpResponse mockResponse = Mockito.mock(CloseableHttpResponse.class);`
I can't set anything as being mocked one but I want to set `statuCode` and `header` dynamically for each test.

resources/sdk/purecloudjava/templates/ApiClient.mustache
PrinceNebulon: I think leaving them fully configurable is ok; however, we're only explicitly documenting the max retry time. Because the SDKs are open source and the retry behavior is optional (as is using the SDK to begin with), blocking consumers from configuring certain options will only serve to cause some users to use a custom build of the SDK or avoid using the logic we offer in favor of implementing their own to work around ours. Less static configurations mean less use cases that prevent people from using the SDK.

resources/sdk/purecloudjava/templates/README.mustache
```diff
@@ -128,6 +128,25 @@ Specify the connector in the builder:
 .withProperty(ApiClientConnectorProperty.CONNECTOR_PROVIDER, new OkHttpClientConnectorProvider())
 ```
 
+#### Setting the max retry time
+
+In default, Java SDK makes a call to API without a Retry option but this feature is configurable.
```
PrinceNebulon: This paragraph could use some rewording:

> By default, the Java SDK does not automatically retry any failed requests. To enable automatic retries, provide a `RetryConfiguration` object with the maximum number of seconds to retry requests when building the `ApiClient` instance.

deerghayu: Updated the docs.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -560,15 +567,18 @@ public class ApiClient implements AutoCloseable {
         ApiClientConnectorRequest connectorRequest = prepareConnectorRequest(request, isAuthRequest);
         ApiClientConnectorResponse connectorResponse = null;
         try {
-            connectorResponse = connector.invoke(connectorRequest);
+            Retry retry = new Retry(retryConfiguration == null ? new RetryConfiguration() : retryConfiguration);
```
chuckpulfer: By default this will new RetryConfiguration() for every request. It might be worth creating a static final default RetryConfiguration and using that instead of creating a new one for each request.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1005,95 @@ public class ApiClient implements AutoCloseable {
         @Override
         public void close() throws Exception { }
     }
+
+    public static class RetryConfiguration {
+        public long backoffInterval = 300000L;
```
chuckpulfer: Mark these private and access them with getters and setters.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1005,95 @@ public class ApiClient implements AutoCloseable {
...
+                throw new IllegalArgumentException("maxRetriesBeforeBackoff can not be a negative integer");
+            }
+            this.maxRetriesBeforeBackoff = maxRetriesBeforeBackoff;
+        }
+    }
+
+    public static class Retry {
+        private long backoffInterval;
+        private long defaultDelay;
+        private int maxRetryTime;
```
chuckpulfer: Add units: maxRetryTimeMs or maxRetryTimeSeconds

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1005,95 @@ public class ApiClient implements AutoCloseable {
...
+        private int maxRetriesBeforeBackoff;
+        private int retryCountBeforeBackOff = 0;
+        private long waitTimeInMillis;
```
chuckpulfer: Since this relates to the Retry-After header I would rename it to retryAfterMs

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1005,95 @@ public class ApiClient implements AutoCloseable {
...
+            if (maxRetriesBeforeBackoff < 0) {
+                throw new IllegalArgumentException("maxRetriesBeforeBackoff can not be a negative integer");
+            }
+            this.maxRetriesBeforeBackoff = maxRetriesBeforeBackoff;
+        }
+    }
+
+    public static class Retry {
+        private long backoffInterval;
+        private long defaultDelay;
```
chuckpulfer: defaultDelay -> retryAfterDefaultMs

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -975,4 +1005,95 @@ public class ApiClient implements AutoCloseable {
...
+        public void setMaxRetriesBeforeBackoff(int maxRetriesBeforeBackoff) {
+            if (maxRetriesBeforeBackoff < 0) {
+                throw new IllegalArgumentException("maxRetriesBeforeBackoff can not be a negative integer");
+            }
+            this.maxRetriesBeforeBackoff = maxRetriesBeforeBackoff;
+        }
+    }
+
+    public static class Retry {
+        private long backoffInterval;
```
chuckpulfer: Add units backoffIntervalSeconds or backoffIntervalMs

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -83,6 +87,9 @@ public class ApiClient implements AutoCloseable {
     private final Map<String, Authentication> authentications;
     private final ApiClientConnector connector;
 
+    private RetryConfiguration retryConfiguration;
```
Mikosz91: This can be final too

deerghayu: True. added final.

resources/sdk/purecloudjava/tests/ApiClientRetryAsyncTest.java
PrinceNebulon: @deerghayu this still needs to be addressed

deerghayu: `CloseableHttpResponse mockResponse = Mockito.mock(CloseableHttpResponse.class);
        mockResponse.setStatusLine(new ProtocolVersion("HTTP/1.1", 1, 1), 429, "Bad GateWay");
        mockResponse.setHeader("Retry-after", "1");`

` doReturn(mockResponse).when(spyClient).execute(any(HttpUriRequest.class));`

If you were suggesting something like this, it won't work since status code and header will not be set as being a mock object. (I'm trying to set different status code and header for each test)

Please correct me if I'm missing something from the comment.

deerghayu: I refactored the mock Response based on the suggestion. Thanks for the comment.

PrinceNebulon:
Not sure what happened, but the SdkTests tests aren't being run when I build the Java SDK from your fork locally. I'm invoking it normally using `node sdkBuilder.js --sdk purecloudjava` and my local config file is identical between the two repos.

Output from `mypurecloud/master`:

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running TestSuite
[TEST] beforeTest
Before test
[SUCCESS] beforeTest
[TEST] traceBasicInformation
Not in PureCloudRegionHosts using string value
PURECLOUD_ENVIRONMENT=inindca.com
PURECLOUD_CLIENT_ID=03b5e96f-ee0c-4fb9-bf7d-d7707a944826
userEmail=82f06128-0bc4-4855-a7fb-bc8adafe3494@inindca.com
[SUCCESS] traceBasicInformation
[TEST] authenticate
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[SUCCESS] authenticate
[TEST] createUser
Created user with ID 0a8d8442-c9c6-4630-960d-317151d0fb79
[SUCCESS] createUser
[TEST] updateUser
[SUCCESS] updateUser
[TEST] setProfileSkills
[SUCCESS] setProfileSkills
[TEST] testNotifications
system presence -> BUSY
system presence -> AVAILABLE
[SUCCESS] testNotifications
[TEST] getUser
[SUCCESS] getUser
[TEST] deleteUser
system presence -> OFFLINE
[SUCCESS] deleteUser
[TEST] afterTest
After test
[SUCCESS] afterTest
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4.779 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

Output from `deerghayu/master`:

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running TestSuite
[TEST] beforeTest
Before test
[SUCCESS] beforeTest
[TEST] setup
[SUCCESS] setup
[TEST] invokeTestWith_429
Not in PureCloudRegionHosts using string value
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[SUCCESS] invokeTestWith_429
[TEST] setup
[SUCCESS] setup
[TEST] invokeTestWith_429_And_No_MaxRetryTime
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeTestWith_429_And_No_MaxRetryTime
[TEST] setup
[SUCCESS] setup
[TEST] invokeTestWith_502
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeTestWith_502
[TEST] setup
[SUCCESS] setup
[TEST] invokeTestWith_503
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeTestWith_503
[TEST] setup
[SUCCESS] setup
[TEST] invokeTestWith_504
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeTestWith_504
[TEST] setup
[SUCCESS] setup
[TEST] invokeTestWith_504_No_MaxRetryTime
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeTestWith_504_No_MaxRetryTime
[TEST] setup
[SUCCESS] setup
[TEST] shouldRetryTestWith_200
[SUCCESS] shouldRetryTestWith_200
[TEST] setup
[SUCCESS] setup
[TEST] shouldRetryTestWith_429
[SUCCESS] shouldRetryTestWith_429
[TEST] setup
[SUCCESS] setup
[TEST] shouldRetryTestWith_502
[SUCCESS] shouldRetryTestWith_502
[TEST] setup
[SUCCESS] setup
[TEST] shouldRetryTestWith_502_And_0_MaxRetryTime
[SUCCESS] shouldRetryTestWith_502_And_0_MaxRetryTime
[TEST] setup
[SUCCESS] setup
[TEST] shouldRetryTestWith_503_And_RetryConfig
[SUCCESS] shouldRetryTestWith_503_And_RetryConfig
[TEST] setup
[SUCCESS] setup
[TEST] shouldRetryTestWith_504_And_No_RetryAfter_Header
[SUCCESS] shouldRetryTestWith_504_And_No_RetryAfter_Header
[TEST] setup
[SUCCESS] setup
[TEST] invokeAsyncTestWith_429
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeAsyncTestWith_429
[TEST] setup
[SUCCESS] setup
[TEST] invokeAsyncTestWith_429_And_No_MaxRetryTime
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeAsyncTestWith_429_And_No_MaxRetryTime
[TEST] setup
[SUCCESS] setup
[TEST] invokeAsyncTestWith_502
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeAsyncTestWith_502
[TEST] setup
[SUCCESS] setup
[TEST] invokeAsyncTestWith_503
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeAsyncTestWith_503
[TEST] setup
[SUCCESS] setup
[TEST] invokeAsyncTestWith_504
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeAsyncTestWith_504
[TEST] setup
[SUCCESS] setup
[TEST] invokeAsyncTestWith_504_And_No_MaxRetryTime
Not in PureCloudRegionHosts using string value
[SUCCESS] invokeAsyncTestWith_504_And_No_MaxRetryTime
[TEST] afterTest
After test
[SUCCESS] afterTest
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 133.998 s - in TestSuite
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0
```

deerghayu:
I updated the testng.xml file to run multiple test classes. 26 tests are passed locally.
 @PrinceNebulon please try running it locally again.

PR #53, opened 2020-03-20
API-4992 Add authenticated proxy support to python sdk

ronanwatkins:
Requested here: https://developer.mypurecloud.com/forum/t/implementing-proxy-with-authentication/6738


PrinceNebulon:
Code looks good, can you also add information on using this to the readme template? I already merged the .NET PR, but you can update that readme in this PR.

PR #66, opened 2020-07-15
DEVCTR-97 Adding token refresh logic to SDKs for Code Authorization flow

ronanwatkins:
Adding token refresh logic to SDKs for Code Authorization flow and adding helper methods for Code Authorization where needed.

### DIFF
--- a/resources/sdk/pureclouddotnet/extensions/Extensions/AuthExtensions.cs
+++ b/resources/sdk/pureclouddotnet/extensions/Extensions/AuthExtensions.cs
@@ -21,20 +21,27 @@ namespace {{=it.packageName}}.Extensions
         /// <param name="redirectUri"></param>
         /// <param name="authorizationCode"></param>
         /// <returns></returns>
-        public static AuthTokenInfo PostToken(this ApiClient apiClient, string clientId, string clientSecret,string redirectUri = "", string authorizationCode = "")
+        public static AuthTokenInfo PostToken(this ApiClient apiClient, string clientId, string clientSecret,string redirectUri = "", string authorizationCode = "", bool isRefreshRequest = false)
         {
-            var response = apiClient.PostTokenWithHttpInfo(clientId, clientSecret, redirectUri, authorizationCode);
+            var response = apiClient.PostTokenWithHttpInfo(clientId, clientSecret, redirectUri, authorizationCode, isRefreshRequest);
             return response.Data;
         }
 
         public static ApiResponse<AuthTokenInfo> PostTokenWithHttpInfo(this ApiClient apiClient, string clientId,
-            string clientSecret, string redirectUri = "", string authorizationCode = "")
+            string clientSecret, string redirectUri = "", string authorizationCode = "", bool isRefreshRequest = false)
         {
             var path_ = "/token";
 
             // This may be uninitialized if no API classes have been constructed yet
             if (apiClient.Configuration == null)
                 apiClient.Configuration = new Configuration(apiClient);
+
+            // If redirectUri is not null this is a Code Authorization grant and we need to save the clientId and clientSecret for a transparent token refresh
+            if (!string.IsNullOrEmpty(redirectUri)) {
+                apiClient.UsingCodeAuth = true;
+                apiClient.ClientId = clientId;
+                apiClient.ClientSecret = clientSecret;
+            }
 
             var pathParams = new Dictionary<String, String>();
             var queryParams = new Dictionary<String, String>();
@@ -60,10 +67,15 @@ namespace {{=it.packageName}}.Extensions
                 headerParams.Add("Accept", httpHeaderAccept);
 
             // Add form params
-            formParams.Add("grant_type",
-                string.IsNullOrEmpty(authorizationCode) ? "client_credentials" : "authorization_code");
-            if (!string.IsNullOrEmpty(authorizationCode))
-                formParams.Add("code", apiClient.ParameterToString(authorizationCode));
+            if (isRefreshRequest) {
+                formParams.Add("grant_type", "refresh_token");
+                formParams.Add("refresh_token", authorizationCode);
+            } else {
+                formParams.Add("grant_type",
+                    string.IsNullOrEmpty(authorizationCode) ? "client_credentials" : "authorization_code");
+                if (!string.IsNullOrEmpty(authorizationCode))
+                    formParams.Add("code", apiClient.ParameterToString(authorizationCode));
+            }
             if (!string.IsNullOrEmpty(redirectUri))
                 formParams.Add("redirect_uri", apiClient.ParameterToString(redirectUri));
 
@@ -85,9 +97,12 @@ namespace {{=it.packageName}}.Extensions
                 throw new ApiException(statusCode, "Error calling PostToken: " + response.ErrorMessage,
                     response.ErrorMessage);
 
+            var authTokenInfo = (AuthTokenInfo) apiClient.Deserialize(response, typeof (AuthTokenInfo));
+            apiClient.Configuration.AuthTokenInfo = authTokenInfo;
+
             return new ApiResponse<AuthTokenInfo>(statusCode,
                 response.Headers.ToDictionary(x => x.Name, x => x.Value.ToString()),
-                (AuthTokenInfo) apiClient.Deserialize(response, typeof (AuthTokenInfo)),
+                authTokenInfo,
                 response.Content,
                 response.StatusDescription);
         }
--- a/resources/sdk/pureclouddotnet/templates/ApiClient.mustache
+++ b/resources/sdk/pureclouddotnet/templates/ApiClient.mustache
@@ -14,6 +14,7 @@ using System.Net;
 using System.Text;
 using Newtonsoft.Json;
 using RestSharp;
+using {{packageName}}.Extensions;
 
 namespace {{packageName}}.Client
 {
@@ -110,6 +111,11 @@ namespace {{packageName}}.Client
         public RetryConfiguration RetryConfig { get; set; }
         private static readonly RetryConfiguration DEFAULT_RETRY_CONFIG = new RetryConfiguration();
 
+        // These fields are only applicable to the Code Authorization OAuth flow:
+        public bool UsingCodeAuth { get; set; }
+        public string ClientId { get; set; }
+        public string ClientSecret { get; set; }
+
         // Creates and sets up a RestRequest prior to a call.
         private RestRequest PrepareRequest(
             String path, RestSharp.Method method, List<Tuple<String, String>> queryParams, Object postBody,
@@ -156,6 +162,36 @@ namespace {{packageName}}.Client
             return request;
         }
 
+        private void HandleExpiredAccessToken()
+        {
+            if (Monitor.TryEnter(Configuration, 0))
+            {
+                try
+                {
+                    Extensions.AuthExtensions.PostToken(this, ClientId, ClientSecret, authorizationCode: Configuration.AuthTokenInfo.RefreshToken, isRefreshRequest: true);
+                }
+                catch (Exception e)
+                {
+                    throw new ApiException(500, e.Message);
+                }
+                finally
+                {
+                    Monitor.Exit(Configuration);
+                }
+            }
+            else
+            {
+                // Abort with error if we have waited the configured time and refresh still isn't complete
+                if (!Monitor.TryEnter(Configuration, TimeSpan.FromSeconds(Configuration.RefreshTokenWaitTime))) {
+                    throw new ApiException(500, $"Token refresh took longer than {Configuration.RefreshTokenWaitTime} seconds");
+                }
+                else
+                {
+                    Monitor.Exit(Configuration);
+                }
+            }
+        }
+
         /// <summary>
         /// Makes the HTTP request (Sync).
         /// </summary>
@@ -200,6 +236,17 @@ namespace {{packageName}}.Client
                response = RestClient.ExecuteTaskAsync(request).Result;
             {{/supportsUWP}}
             }while(retry.ShouldRetry(response));
+
+            if (UsingCodeAuth && Configuration.ShouldRefreshAccessToken)
+            {
+                int statusCode = (int) response.StatusCode;
+                if (statusCode == 401)
+                {
+                    HandleExpiredAccessToken();
+                    return CallApi(path, method, queryParams, postBody, headerParams, formParams, fileParams, pathParams, contentType);
+                }
+            }
+
             return (Object) response;
         }
         {{#supportsAsync}}
@@ -232,6 +279,16 @@ namespace {{packageName}}.Client
             {
                 response = await RestClient.ExecuteTaskAsync(request);
             }while(retry.ShouldRetry(response));
+
+            if (UsingCodeAuth && Configuration.ShouldRefreshAccessToken)
+            {
+                int statusCode = (int) response.StatusCode;
+                if (statusCode == 401)
+                {
+                    HandleExpiredAccessToken();
+                    return await CallApiAsync(path, method, queryParams, postBody, headerParams, formParams, fileParams, pathParams, contentType);
+                }
+            }
 
             return (Object)response;
         }{{/supportsAsync}}
--- a/resources/sdk/pureclouddotnet/templates/README.mustache
+++ b/resources/sdk/pureclouddotnet/templates/README.mustache
@@ -73,7 +73,30 @@ var accessTokenInfo = Configuration.Default.ApiClient.PostToken("18a4c365-7ea3-4
   "http://redirecturi.com/",
   "6Zxcb0oASMBI55wQJ6bVmOmO57k8CxXBKgzDKtYXbtk");
 Console.WriteLine("Access token=" + accessTokenInfo.AccessToken);
-Configuration.Default.AccessToken = accessTokenInfo.AccessToken;
+```
+
+By default the SDK will transparently request a new access token using the refresh token when the access token expires. If you wish to implement the refresh logic set _ShouldRefreshAccessToken_ to false and store the refresh token from the auth response:
+
+```{"language":"csharp"}
+var refreshToken = accessTokenInfo.RefreshToken;
+Configuration.Default.ShouldRefreshAccessToken = false;
+```
+
+You can use the _ExpiresIn_ value to determine how long the token will live and proactively request a new one before it expires.
+
+```{"language":"csharp"}
+var tokenTimeToLive = authTokenInfo.ExpiresIn;
+```
+
+When the access token expires refresh it using the _PostToken_ method using the same clientId and clientSecret as used to request it.
+
+```{"language":"csharp"}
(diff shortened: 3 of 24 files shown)


resources/sdk/pureclouddotnet/extensions/Extensions/AuthExtensions.cs
```diff
@@ -85,9 +97,18 @@ public static AuthTokenInfo PostToken(this ApiClient apiClient, string clientId,
                 throw new ApiException(statusCode, "Error calling PostToken: " + response.ErrorMessage,
                     response.ErrorMessage);
 
+            var authTokenInfo = (AuthTokenInfo) apiClient.Deserialize(response, typeof (AuthTokenInfo));
+
+            if (apiClient.UsingCodeAuth) {
```
PrinceNebulon, line 102: Let's remove all the code in this if block. The .NET SDK does not mutate the ApiClient instance when performing authorization (others do, so this feedback only applies here). The consuming application is expected to keep track of the refresh token and credentials. There's a few lines above (line 40) that should be removed as well. Instead, use a local variable to track if it's a code auth request.

PrinceNebulon: I just realized why you did this; the implementation in ApiClient to automatically get a refresh token requires it. Because of that, let's change this to store the auth response in all cases, not just for code auth grants. I'd suggest removing the individual properties and just adding an `AuthTokenInfo` property.

resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -156,6 +169,44 @@ namespace {{packageName}}.Client
...
+        }
+
+        private void HandleExpiredAccessToken()
+        {
+            if (Monitor.TryEnter(RefreshTokenLock, 0))
+            {
+                try
+                {
+                    AuthTokenInfo authTokenInfo = Extensions.AuthExtensions.PostToken(this, ClientId, ClientSecret, authorizationCode: RefreshToken, isRefreshRequest: true);
+                    Configuration.Default.AccessToken = authTokenInfo.AccessToken;
```
PrinceNebulon, line 186: `Configuration.Default.AccessToken` is incorrect here. That's accessing the singleton default instance of Configuration, which may be different than the instance being used by this instance of ApiClient. I believe it should use `this.Configuration.AccessToken`.

resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -156,6 +169,44 @@ namespace {{packageName}}.Client
             return request;
         }
 
+        private bool HasAccessTokenExpired(int statusCode)
```
PrinceNebulon, line 172: This method shouldn't be necessary. It doesn't matter if there _should_ be time left on the token; it only matters if the token was invalid. In the couple places this method is invoked, just check to see if the status code is 401 and invoke `HandleExpiredAccessToken()` if it is.

resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -156,6 +169,44 @@ namespace {{packageName}}.Client
...
+        {
+            TimeSpan timeSinceEpoch = (DateTime.UtcNow - new DateTime(1970, 1, 1));
+            long timeElapsed = (long) (timeSinceEpoch.TotalSeconds) - TokenIssueTime;
+            return statusCode == 401 && timeElapsed > TokenTimeToLive;
+        }
+
+        private void HandleExpiredAccessToken()
+        {
+            if (Monitor.TryEnter(RefreshTokenLock, 0))
```
PrinceNebulon, line 181: The object for locking may not be 100% correct here. Because the access token is stored on the `Configuration` instance, it's possible for two instances of `ApiClient` to use the same `Configuration` instance, which could result in multiple `ApiClient` instances attempting to refresh the same access token simultaneously; one will succeed and one will fail putting it in a bad state. If you use the Configuration object, it should prevent this.

resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -156,6 +169,44 @@ namespace {{packageName}}.Client
...
+                }
+                finally
+                {
+                    Monitor.Exit(RefreshTokenLock);
+                }
+            }
+            else
+            {
+                // Abort with error if we have waited 2 seconds and refresh still isn't complete
+                if (!Monitor.TryEnter(RefreshTokenLock, TimeSpan.FromSeconds(2))) {
```
PrinceNebulon, line 200: Make this duration configurable and default to 10 seconds. For applications using a slow internet connection or if auth service is a little backed up, this could take longer.

resources/sdk/pureclouddotnet/templates/README.mustache
```diff
@@ -76,6 +76,31 @@ Console.WriteLine("Access token=" + accessTokenInfo.AccessToken);
...
+When the access token expires refresh it using the _PostToken_ method using the same clientId and clientSecret as used to request it.
+
+```{"language":"csharp"}
+var accessTokenInfo = Configuration.Default.ApiClient.PostToken("18a4c365-7ea3-4f0g-9fb7-884fb4d2e9c6",
+  "M7FfdYQyL5TA6BdbEZ8M9-Wx4uZai1rNQ7jcuFdcJJo",
+  authorizationCode: refreshToken,
+  isRefreshRequest: true);
+Console.WriteLine("Access token=" + accessTokenInfo.AccessToken);
+refreshToken = accessTokenInfo.RefreshToken;
+Configuration.Default.AccessToken = accessTokenInfo.AccessToken;
```
PrinceNebulon, line 101: This line shouldn't be necessary anymore if PostToken is mutating the configuration and setting the auth info now.

PrinceNebulon:
I've reviewed the .NET SDK so far. It looks like some of the feedback applies to other SDKs, so I'll hold off on reviewing them for now. Let me know if you want to hop on a zoom session tomorrow to discuss.

resources/sdk/pureclouddotnet/templates/api_doc.mustache
```diff
@@ -51,7 +51,11 @@ namespace Example
             // Uncomment below to setup prefix (e.g. BEARER) for API key, if needed
             // Configuration.Default.ApiKeyPrefix.Add('{{{keyParamName}}}', 'BEARER');{{/isApiKey}}{{#isOAuth}}
             // Configure OAuth2 access token for authorization: {{{name}}}
-            Configuration.Default.AccessToken = "YOUR_ACCESS_TOKEN";{{/isOAuth}}{{/authMethods}}{{/hasAuthMethods}}
+            // The following example is using the [Authorization Code Grant](https://developer.mypurecloud.com/api/rest/authorization/use-authorization-code.html)
```
PrinceNebulon, line 54: Including the link to the docs is a nice touch, but this is in a code block so markdown won't be rendered here. It should just be plain text.

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -549,7 +607,8 @@ public class ApiClient implements AutoCloseable {
         String reasonPhrase = response.getStatusReasonPhrase();
         Map<String, String> headers = response.getHeaders();
 
-        if (statusCode >= 200 && statusCode < 300) {
+        // If the response indicates that the access_token has expired, and we wish to refresh it transparently, we'll return from this method and let the caller handle it
+        if (statusCode >= 200 && statusCode < 300 || statusCode == 401 && shouldRefreshAccessToken) {
```
PrinceNebulon, line 611: How does this work? It appears that it's going to process the response as if it were successful when the response is 401, which should fail because it's going to parse the response expecting it to the success response type, but the actual response is an error body.

resources/sdk/purecloudjava/templates/README.mustache
```diff
@@ -77,6 +77,42 @@ System.out.println("Authentication successful. Access token expires in " + authR
 
 ```
 
+See example on how to authenticate with an authorization code below. For more information see the article on [Code Authorization](https://developer.mypurecloud.com/api/rest/authorization/use-authorization-code.html)
```
PrinceNebulon, line 80: Shouldn't this statement should go a few lines down under the auth code heading?

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -727,122 +866,134 @@ class ApiClient {
 	 * @returns {Promise} A Promise object.
 	 */
 	callApi(path, httpMethod, pathParams, queryParams, headerParams, formParams, bodyParam, authNames, contentTypes, accepts) {
-		var url = this.buildUrl(path, pathParams);
-		var request = superagent(httpMethod, url);
-
-		if (this.proxy && request.proxy) {
-			request.proxy(this.proxy);
-		}
+		return new Promise((resolve, reject) => {
```
PrinceNebulon, line 869: Can you summarize the changes to the `callApi` function? It looks like maybe you rearranged some parts to properly handle refreshing the token, but the diff is making it look like there were a lot of changes.

resources/sdk/purecloudkotlin/templates/ApiClient.kt.mustache
```diff
@@ -389,7 +446,8 @@ class ApiClient private constructor(builder: Builder) : AutoCloseable {
         val statusCode = response!!.statusCode
         val reasonPhrase = response.statusReasonPhrase
         val headers = response.getHeaders()
-        return if (statusCode in 200..299) {
+        // If the response indicates that the access_token has expired, and we wish to refresh it transparently, we'll return from this method and let the caller handle it
+        return if (statusCode in 200..299 || shouldRefreshAccessToken && statusCode == 401) {
```
PrinceNebulon, line 450: Why is the 401 response being handled as a success here? Same concerns as this condition mentioned in another comment.

resources/sdk/purecloudkotlin/templates/ApiClient.kt.mustache
```diff
@@ -409,6 +467,36 @@ class ApiClient private constructor(builder: Builder) : AutoCloseable {
...
+            var timeSpentWaiting = 0.0
+            val sleepDuration = 200
+            // Sleep for 200ms at a time for a maximum of refreshTokenWaitTime for other thread to complete refresh
+            while (timeSpentWaiting < refreshTokenWaitTime) {
+                try {
+                    Thread.sleep(sleepDuration.toLong())
+                } catch (e: InterruptedException) {
+                    Thread.currentThread().interrupt()
+                }
+                timeSpentWaiting += sleepDuration.toDouble() / 1000
```
PrinceNebulon, line 493: Adding sleeps together like this will miss some time and the total duration waited will be longer than calculated as it does not take into account processing time. Instead, set a variable with the current time before the loop and make the while condition something like `currentTime - startTime < refreshTokenWaitTime`. Using unix timestamps is one good option because using a single number to represent the timestamp makes calculating the duration very easy.

resources/sdk/purecloudkotlin/templates/README.mustache
```diff
@@ -45,7 +45,7 @@ val users = usersApi.getUsers(null, null, null, null, null, null, null)
 
 For user applications, the consuming application must complete an implicit, auth token, or SAML2 Bearer OAuth flow to get an access token outside the scope of the SDK. Once an access token is obtained, it should be set on the SDK via constructing a new ApiClient instance (use `withAccessToken(token: String)`). For more information about authenticating with OAuth, see the Developer Center article [Authorization](https://developer.mypurecloud.com/api/rest/authorization/index.html). For more information about SAML2 Bearer Oauth flow view the example below 
 
-## Authentication with SAML2Bearer token
+#### Authentication with SAML2Bearer token
 
 ```{"language":"kotlin"}
 val clientId = System.getenv("PURECLOUD_CLIENT_ID")
```
PrinceNebulon, line 51: While you're making these edits, can you rename the env vars in the readme files to use `GENESYS_CLOUD`? e.g.  `GENESYS_CLOUD_CLIENT_ID`

resources/sdk/purecloudkotlin/templates/README.mustache
```diff
@@ -59,7 +59,44 @@ val region = PureCloudRegionHosts.us_east_1
 val apiClient: ApiClient = ApiClient.Builder.standard().withBasePath(region).build()
 val authResponse = apiClient.authorizeSaml2Bearer(clientId, clientSecret, orgName, encodedSamlAssertion)
 println("Authentication successful. Access token expires in ${authResponse.getBody()?.expires_in} seconds")
+```
+
+See example on how to authenticate with an authorization code below. For more information see the article on [Code Authorization](https://developer.mypurecloud.com/api/rest/authorization/use-authorization-code.html)
```
PrinceNebulon, line 64: Should this statement be under the auth code heading below?

resources/sdk/purecloudpython/templates/api_client.mustache
```diff
@@ -173,11 +260,35 @@ class ApiClient(object):
...
+                self.refresh_in_progress = False
+                self.refresh_token_lock.release()
+        else:
+            time_spent_waiting = 0.0
+            sleep_duration = 0.200
+            # Wait maximum of refresh_token_wait_time seconds for other thread to complete refresh
+            # It would be ideal to use refresh_token_lock.acquire with a timeout value here but that was added in python 3.2 and we have to support older versions
+            while time_spent_waiting < Configuration().refresh_token_wait_time:
+                time.sleep(sleep_duration)
+                time_spent_waiting += sleep_duration
```
PrinceNebulon, line 278: Same comment as kotlin about measuring sleep times.

PrinceNebulon:
@carnellj would appreciate a second set of eyes when you have a chance

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
ronanwatkins: For the purpose of calling the function recursively after refreshing the access token I created a nested function named "sendRequest" inside callApi.
sendRequest retains most of the functionality of callApi, references to "this" have been changed to "that" which is partially why the diff looks so different, the other reason being everything is nested a further level now.
The functionality for handling an expired access token has been added at line 970. Previously it called "reject(data);" when an error occurred, now it calls _handleExpiredAccessToken and recursively calls itself if a 401 code is received and shouldRefreshAccessToken is set. It will call "reject(data);" if the status code isn't 401, if shouldRefreshAccessToken is not set or an error occurred in _handleExpiredAccessToken.

resources/sdk/purecloudjava/templates/ApiClient.mustache
ronanwatkins: That probably wasn't the cleanest solution. I'll remove the changes to this method and update the callers (getAPIResponse and getAPIResponseAsyncWithRetry) to handle the exception. They can request a new access token if the status code is 401 and shouldRefreshAccessToken is set, or rethrow the exception if those conditions aren't met

resources/sdk/pureclouddotnet/templates/README.mustache
```diff
@@ -73,7 +73,30 @@ var accessTokenInfo = Configuration.Default.ApiClient.PostToken("18a4c365-7ea3-4
...
+```
+
+By default the SDK will transparently request a new access token using the refresh token when the access token expires. If you wish to implement the refresh logic set _ShouldRefreshAccessToken_ to false and store the refresh token from the auth response:
+
+```{"language":"csharp"}
+var refreshToken = accessTokenInfo.RefreshToken;
+Configuration.Default.ShouldRefreshAccessToken = false;
+```
+
+You can use the _ExpiresIn_ value to determine how long the token will live and proactively request a new one before it expires.
```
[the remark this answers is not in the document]
ronanwatkins: What I meant here was the user can store the expires in value and do something such as setting a function to execute a few minutes before the token expires to proactively request a new access token.

resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -144,6 +145,14 @@ func (c *APIClient) CallAPI(path string, method string,
 	// Read body
 	body, _ := ioutil.ReadAll(res.Body)
 
+	if res.StatusCode == 401 && c.configuration.ShouldRefreshAccessToken {
```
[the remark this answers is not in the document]
ronanwatkins: Good spot, I'll update it

resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -198,6 +207,28 @@ func (c *APIClient) ParameterToString(obj interface{}, collectionFormat string)
...
+		defer atomic.StoreInt64(&c.configuration.RefreshInProgress, 0)
+		_, err := c.configuration.RefreshAuthorizationCodeGrant(c.configuration.ClientID, c.configuration.ClientSecret, c.configuration.RefreshToken)
+		return err
+	} else {
+		// Wait maximum of RefreshTokenWaitTime seconds for other thread to complete refresh
+		timeSpentWaiting := float32(0.0)
+		sleepDuration := time.Millisecond * 200
+		for int(timeSpentWaiting) < c.configuration.RefreshTokenWaitTime {
+			time.Sleep(sleepDuration)
+			timeSpentWaiting += float32(sleepDuration) / 1000000 / 1000
```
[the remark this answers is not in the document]
ronanwatkins: My logic there wasn't great. I'll update it and bring it in line with the logic in the other SDKs and comment where necessary.

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -385,6 +390,140 @@ class ApiClient {
...
+	 */
+	_handleExpiredAccessToken() {
+		return new Promise((resolve, reject) => {
+			if (typeof window !== 'undefined') {
+				reject(new Error('This method is not supported in a browser.'));
+				return;
+			}
+			if (!this.refreshInProgress) {
+				this.refreshInProgress = true;
+				this.refreshCodeAuthorizationGrant(this.clientId, this.clientSecret, this.authData.refreshToken)
```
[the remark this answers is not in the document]
ronanwatkins: Can't use that outside of an async function unfortunately.

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -385,6 +390,140 @@ class ApiClient {
...
+	 * @param {string} redirectUri - Authorized redirect URI for your Code Authorization client
+	 */
+    refreshCodeAuthorizationGrant(clientId, clientSecret, refreshToken) {
+		return new Promise((resolve, reject) => {
+			if (typeof window !== 'undefined') {
+				reject(new Error('The Code Authorization grant is not supported in a browser.'));
+				return;
+			}
+			var encodedData = new Buffer(clientId + ':' + clientSecret).toString('base64');
+			var request = superagent('POST', `https://login.${this.environment}/oauth/token`);
```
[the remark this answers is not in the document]
ronanwatkins: Done. Created two utility functions to handle CodeAuth stuff.

timsmithgenesys (approved):

PR #73, opened 2020-08-19
Fixes needed for go build

ronanwatkins:
Adding extra parameter to "usersAPI.GetUsers" in test code
Ignoring empty strings when passed as API parameters [potential temporary fix]


PrinceNebulon:
Holding off on merging this pending resolution of API-5550. This shouldn't be needed once that is fixed.

PrinceNebulon (approved):

PR #75, opened 2020-09-15
Adding fix for ENOBUFS error building dotnet SDK

ronanwatkins:


resources/sdk/pureclouddotnet/scripts/prebuild-postrun.js
```diff
@@ -50,7 +50,8 @@ try {
 
 	console.log('downloading nuget...');
 	let data = cp.execFileSync('curl', ['--silent', '-L', 'https://dist.nuget.org/win-x86-commandline/latest/nuget.exe'], {
-		encoding: 'binary'
+		encoding: 'binary',
+		maxBuffer: 1024 * 1024 * 10
```
PrinceNebulon, line 54: Why has this worked for years for multiple people, and still works except for one person?

ronanwatkins: The only reason I can think of is the fact that Hakan's and my macBook are new and have more recent OS versions. There must be some difference causing more bytes to be outputted on stdout

PrinceNebulon (approved):

PR #80, opened 2020-11-13
DEVENGAGE-338:  Cleaned up the .NET conversion code to make sure that…

carnellj:
… we are properly serializing date only fields for WEM

### DIFF
--- a/resources/sdk/pureclouddotnet/extensions/Model/LocalDate.cs
+++ b/resources/sdk/pureclouddotnet/extensions/Model/LocalDate.cs
@@ -1,0 +1,26 @@
+using System;
+using {{= it.packageName}}.Model;
+
+namespace {{=it.packageName }}.Model
+{
+  /// <summary>
+  /// The LocalDate class is a wrapper class that maps the Genesys Cloud swagger definition to a Date only class.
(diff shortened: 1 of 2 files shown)


resources/sdk/pureclouddotnet/extensions/Model/LocalDate.cs
```diff
@@ -0,0 +1,26 @@
...
+  /// The LocalDate class is a wrapper class that maps the Genesys Cloud swagger definition to a Date only class.
+  /// This class written to deal with DEVENGAGE-338 where a customer raised an issue where they were getting improperly
+  /// formatted Dates (e.g. with a timestamp) because the default serialization for a LocalDate was to serialize to a 
+  /// DateTime.
+  /// </summary>  
+public class LocalDate
+{
+  private DateTime dateTime;
+
+  public LocalDate(DateTime dateTime)
```
timsmithgenesys, line 16: I believe this needs a default constructor and the dateTime property needs to be public so this class can be used for deserializing responses from the API as well.

carnellj:
Talked with Tim we are going to close this.  It is going to be a major headache if we want to deserialize back so we are going to map to a string.

PR #83, opened 2020-12-09
DEVENGAGE-287 webmessaging sdk

ronanwatkins:
Configuration for the Java Web Messaging SDK


resources/sdk/webmessagingjava/extensions/ConsoleColors.java
```diff
@@ -0,0 +1,81 @@
+package cloud.genesys.webmessaging.sdk;
+
+// Inspired by https://stackoverflow.com/a/45444716/1124338
+public class ConsoleColors {
```
timsmithgenesys, line 4: This class isn't/shouldn't be used anywhere and should be removed. I only had it in there originally because I was using it during development to aid local logging and debugging. The SDK should use the slf4j logger to emit trace messages.

resources/sdk/webmessagingjava/extensions/extensions/AuthResponse.java
```diff
@@ -0,0 +1,51 @@
+package cloud.genesys.webmessaging.sdk.extensions;
+
+import java.io.Serializable;
+
+public class AuthResponse implements Serializable {
```
timsmithgenesys, line 5: This class shouldn't be needed; IIRC it's only used for handing the response for auth code and client credential token requests.

resources/sdk/webmessagingjava/scripts/mavenDeploy.sh
```diff
@@ -0,0 +1,53 @@
...
+	DPGP_PASSPHRASE="-Dgpg.passphrase=$DPGP_PASSPHRASE"
+fi
+
+# Add maven to PATH
+export PATH=$PATH:/usr/local/maven/bin
+
+# CD to build dir
+cd $BUILD_DIR
+
+echo "RONAN finished"
```
timsmithgenesys, line 42: image

resources/sdk/webmessagingjava/scripts/prebuild-postrun.js
```diff
@@ -0,0 +1,30 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+try {
+	var swaggerCodegenConfigFilePath = process.argv[2];
+	// var version = require(path.resolve(process.argv[3]));
```
timsmithgenesys, line 6: Need to reenable this.

resources/sdk/webmessagingjava/templates/GenesysCloudRegionHosts.mustache
```diff
@@ -0,0 +1,23 @@
+package {{invokerPackage}};
+
+public enum GenesysCloudRegionHosts {
```
timsmithgenesys, line 3: Is it feasible to genericize this and combine it with the websocket hosts file? So these hosts would only define "usw2.pure.cloud" and anything consuming the region setting would use format the URL appropriately depending on if it's being used for the API or a websocket. That would allow the consumer to specify the region once and the SDK can use it everywhere it applies.

ronanwatkins: I've removed GenesysCloudRegionHosts and added some utility methods to GenesysCloudRegionWebSocketHosts format to an API URL.

PR #85, opened 2021-01-05
DEVENGAGE-355: Autogeneration of CLI SDK

ronanwatkins:


resources/sdk/clisdkclient/config.json
```diff
@@ -0,0 +1,111 @@
+{
+  "name": "GO CLI SDK",
+  "envVars": {
+  },
+  "settings": {
+    "sdkRepo": {
+      "repo": "https://github.com/ronanwatkins/platform-client-sdk-go.git",
```
ronanwatkins, line 7: Will need to update this to the repo we're going to use for this SDK

resources/sdk/clisdkclient/extensions/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+
+func SaveConfig(config Configuration) error {
+	viper.Set(fmt.Sprintf("%s.client_credentials", config.ProfileName()), config.ClientID())
+	viper.Set(fmt.Sprintf("%s.client_secret", config.ProfileName()), config.ClientSecret())
+	viper.Set(fmt.Sprintf("%s.environment", config.ProfileName()), config.Environment())
+	//Checking to see if the file does not exist.  It it doesnt we write out the config as default config.toml
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			homeDir, _ := os.UserHomeDir()
+			gcDir := fmt.Sprintf("%s/.gc", homeDir)
```
ronanwatkins, line 76: @carnellj I saw that this failed if the ~/.gc directory wasn't created so I've added this step to create it.

resources/sdk/clisdkclient/extensions/services/commandservice.go
```diff
@@ -0,0 +1,180 @@
...
+
+//CommandService holds the method signatures for all common Command object invocations
+type CommandService interface {
+	Get(uri string) (string, error)
+	List(uri string) (string, error)
+	Post(uri string, payload string) (string, error)
+	Patch(uri string, payload string) (string, error)
+	Put(uri string, payload string) (string, error)
+	Delete(uri string) (string, error)
+	DetermineAction(httpMethod string, operationId string, uri string) (string, error)
```
ronanwatkins, line 27: @carnellj  I've added DetermineAction to this interface to clean up the API code. Also I changed Delete to also return the body so we can see the output of it.

resources/sdk/clisdkclient/extensions/services/commandservice.go
```diff
@@ -0,0 +1,180 @@
...
+	//Appends the individual records from Entities into the array
+	for _, val := range firstPage.Entities {
+		totalResults = append(totalResults, string(val))
+	}
+
+    //Looks up the rest of the pages
+	if firstPage.PageCount > 1 {
+		pagedURI := uri
+		for x := 2; x <= firstPage.PageCount; x++ {
+			if strings.Contains(pagedURI,"pageNumber=") {
```
ronanwatkins, line 87: @carnellj Some changes here for the query string because sometimes the pageNumber will already be part of the URI passed into this method.

resources/sdk/clisdkclient/templates/Makefile.mustache
```diff
@@ -0,0 +1,33 @@
+ENV=GOPRIVATE=bitbucket.org/inindca
```
ronanwatkins, line 1: @carnellj I changed the Makefile to build for the 3 different supported OS's

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -0,0 +1,192 @@
...
+func (r *RESTClient) Delete(uri string) (string, error) {
+	return r.callAPI(http.MethodDelete, uri, "")
+}
+
+func (r *RESTClient) callAPI(method string, uri string, data string) (string, error) {
+	var apiURI *url.URL
+	if !strings.HasPrefix(uri, "/") {
+		uri = fmt.Sprintf("/%v", uri)
+	}
+	if !strings.Contains(r.environment, "localhost") {
```
ronanwatkins, line 81: @carnellj I added support for running against a locally hosted API here and made a small change to the Login function also.

resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -0,0 +1,76 @@
...
+func init() {
+	cobra.OnInitialize()
+	initViperConfig()
+
+	rootCmd.PersistentFlags().StringVarP(&profileName, "profile", "p", "DEFAULT", "Name of the profile to use for configuring the cli")
+
+	rootCmd.AddCommand(versionCmd)
+	{{=it.addCommands}}
+
+	if os.Getenv("GenerateGcDocs") != "" {
```
ronanwatkins, line 64: @carnellj This is where the markdown logs are generated if the "GenerateGcDocs" flag is set

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -0,0 +1,192 @@
...
+
+// HTTPClient interface
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+type HttpStatusError struct {
+	Verb       string
+	Path       string
+	Body       string
```
ronanwatkins, line 31: @carnellj I added the Body to this struct so we can see the reason for the error

resources/sdk/clisdkclient/extensions/cmd/edges/rebootedge.go
```diff
@@ -24,7 +25,8 @@ var rebootEdgesCmd = &cobra.Command{
 		path := "/api/v2/telephony/providers/edges/{edgeId}/reboot"
 		path = strings.Replace(path, "{edgeId}", args[0], 1)
 
-		results, err := CommandService.Post(path, data)
+		retryFunc := services.RetryWithData(path, data, CommandService.Post)
+		results, err := retryFunc(nil)
```
ronanwatkins, line 29: @carnellj The retry configuration goes to default values if it's nil. We can update this to be read from the config file or configured individually in future as required

resources/sdk/clisdkclient/config.json
timsmithgenesys: https://github.com/MyPureCloud/platform-client-sdk-cli

Repo is created, but I haven't added it to the access control list. Go ahead and add it and add yourself as a maintainer so you can do the initial setup in the repo.

resources/sdk/clisdkclient/extensions/cmd/edges/rebootedge.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	edgesCmd.AddCommand(rebootEdgesCmd)
+}
+
+var rebootEdgesCmd = &cobra.Command{
+	Use:   "reboot [edge id]",
+	Short: "reboots a edge in Genesys Cloud",
```
timsmithgenesys, line 18: _an_ edge

resources/sdk/clisdkclient/extensions/cmd/notifications/subscriptions/deletesubscriptions.go
```diff
@@ -0,0 +1,28 @@
...
+	"gc/utils"
+	"gc/services"
+
+	"github.com/spf13/cobra"
+)
+
+var deleteSubscriptionsCmd = &cobra.Command{
+	Use:   "delete [channel id]",
+	Short: "Deletes a subscription by Genesys Cloud channel id",
+	Long:  `Deletes a subscription by skill by Genesys Cloud id`,
```
timsmithgenesys, line 15: The long description references skills. Bad copy/paste?

resources/sdk/clisdkclient/extensions/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+	return &configuration{profileName: profileName,
+		clientID:     viper.GetString(fmt.Sprintf("%s.client_credentials", profileName)),
+		clientSecret: viper.GetString(fmt.Sprintf("%s.client_secret", profileName)),
+		environment:  viper.GetString(fmt.Sprintf("%s.environment", profileName)),
+	}, nil
+}
+
+func SaveConfig(config Configuration) error {
+	viper.Set(fmt.Sprintf("%s.client_credentials", config.ProfileName()), config.ClientID())
+	viper.Set(fmt.Sprintf("%s.client_secret", config.ProfileName()), config.ClientSecret())
```
timsmithgenesys, line 70: @carnellj Should we be obfuscating the client secret in some way so it's not stored in plain text?

resources/sdk/clisdkclient/templates/README.mustache
```diff
@@ -0,0 +1,119 @@
+---
+title: Platform API CLI SDK
+---
```
timsmithgenesys, line 3: add `ispreview: true` to the frontmatter to add the standard "pre-release stuff can change" message to the page when it's rendered in the dev center. The statement below is fine to keep since this file is used in the github repo as well.

resources/sdk/clisdkclient/templates/README.mustache
```diff
@@ -0,0 +1,119 @@
...
+- skills
+- stations 
+- users
+
+In addition, a few objects provide additional capabilities, including:
+- edges - Rebooting Edges
+- phones - Rebooting Phones
+- queues - Get estimated weight time
+- queues - Get users assigned to queue
+- notifications - Manager and listen to notifications from the event service
```
timsmithgenesys, line 32: manager -> manage

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -0,0 +1,197 @@
...
+		Method: strings.ToUpper((method)),
+		Header: make(map[string][]string),
+	}
+
+	//Setting up the auth header
+	request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", r.token))
+	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("Cache-Control", "no-cache")
+
+    //User-Agent and SDK version headers
```
timsmithgenesys, line 100: Indentation is off on these couple lines.

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -0,0 +1,197 @@
...
+	defer resp.Body.Close()
+
+	response, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	responseData := string(response)
+
+	if (resp.StatusCode != http.StatusOK) && (resp.StatusCode != http.StatusAccepted) {
```
timsmithgenesys, line 123: Should this check for any 2xx status code? Maybe do a check like `if resp.StatusCode < 200 || resp.StatusCode >=300`. If not, it should at least also check for StatusCreated (201) and StatusNoContent (204) as those are both used in the API.

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -0,0 +1,197 @@
...
+		httpError := HttpStatusError{Verb: method, Path: uri, StatusCode: resp.StatusCode, Headers: resp.Header, Body: fmt.Sprintf("%s", pretty.Pretty([]byte(responseData)))}
+		return "", httpError
+	}
+
+	return responseData, nil
+}
+
+//Login authenticates the user using the client credentials in their profile.
+func Login(config config.Configuration) (oAuthToken, error) {
```
timsmithgenesys, line 132: Can we call this `Authorize` instead of `Login`? Regardless of what name we choose for the function, the comment should use the word "authorizes" instead of "authenticates". It's pedantic, but authorization describes the act of exchanging credentials for an auth token. Authentication, which describes the act of validating a set of credentials, is only done internally within the auth service.

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	//Setting up the basic auth headers for the call
+	authHeaderString := fmt.Sprintf("%s:%s", config.ClientID(), config.ClientSecret())
+	authHeader := base64.StdEncoding.EncodeToString([]byte(authHeaderString))
+	request.Header.Set("Authorization", fmt.Sprintf("Basic %s", authHeader))
+	request.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	//User-Agent and SDK version headers
+    request.Header.Set("User-Agent", "{{#httpUserAgent}}{{.}}{{/httpUserAgent}}{{^httpUserAgent}}Swagger-Codegen{{/httpUserAgent}}/go-cli")
```
timsmithgenesys, line 153: indentation

timsmithgenesys:
@carnellj @ronanwatkins As a general concept, we should have a test for each handwritten command that we include in lieu of using a generated command for the resource. These tests should verify that the swagger definition for the operation is fully and accurately represented in the handwritten command. This includes checking the description, the names and casing of the path and query parameters, and anything else that is possible to check from the swagger definition. This will be necessary to ensure that the SDK stays up to date with the API's definition as changes are made to API resources where we won't be in the loop to be notified of the change prior to it happening. At least the test can fail the build so we know an update to the handwritten commands is required.

resources/sdk/clisdkclient/extensions/config/config.go
carnellj: I am good with that

resources/sdk/clisdkclient/extensions/services/commandservice.go
carnellj: Sounds good

resources/sdk/clisdkclient/templates/root.mustache
carnellj: Cool

resources/sdk/clisdkclient/templates/restclient.mustache
carnellj: Cool

resources/sdk/clisdkclient/extensions/config/config.go
carnellj: Yes and no.  The problem is that if we obfuscate the code then that means the client must set up all-new credentials only through the Command Line tools.  If the customer ever wants to use this in the context of CI/CD, they would need to manually generate the profiles with the tool and then take the config file and check it into a secret's value or a filesystem.

I am going to chat with info security about this, but frankly, this all runs on the customer's machine and they have complete responsibility to secure these credentials.  We would force obfuscation on any credentials created or used.  It's a good point to at least look into.

resources/sdk/clisdkclient/templates/README.mustache
carnellj: I am good

resources/sdk/clisdkclient/templates/restclient.mustache
carnellj: I think thats a good idea.  When I originally wrote the good I wrote because I only ever dealt with 200 and 202s.  I think we should with your first comment Tim.  If our APIs support other "Success" status code then we should implement this.

resources/sdk/clisdkclient/templates/restclient.mustache
carnellj: I am good with that.  The problem with first cuts of names is that they are often not thought through and I like calling this authorize

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -30,9 +30,10 @@ type HttpStatusError struct {
 	Path       string
 	Body       string
 	StatusCode int
+	Headers    map[string][]string
 }
 
-func (e *HttpStatusError) Error() string {
+func (e HttpStatusError) Error() string {
```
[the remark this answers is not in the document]
carnellj: I am good

resources/sdk/clisdkclient/extensions/cmd/edges/rebootedge.go
carnellj: Hi Ronan,

 I liked your implementation exception for one thing.  I would break the RetryFunctions into a separate package.  We are going to handle other Error and Retry handling strategies in the future and I don't want it tied to the services package.  Traditionally, the services package should be for business services code.  This is a support set of functions and it feels these functions could be in their own practice.

resources/sdk/clisdkclient/extensions/cmd/notifications/subscriptions/deletesubscriptions.go
carnellj: This was a bad copy paste

carnellj:
> @carnellj @ronanwatkins As a general concept, we should have a test for each handwritten command that we include in lieu of using a generated command for the resource. These tests should verify that the swagger definition for the operation is fully and accurately represented in the handwritten command. This includes checking the description, the names and casing of the path and query parameters, and anything else that is possible to check from the swagger definition. This will be necessary to ensure that the SDK stays up to date with the API's definition as changes are made to API resources where we won't be in the loop to be notified of the change prior to it happening. At least the test can fail the build so we know an update to the handwritten commands is required.

I agree with Tim on this.  These should be unit tests and only test the flows of the functionality.  Also keep in mind too Hakan is always working on platform tests.

resources/sdk/clisdkclient/extensions/config/config.go
timsmithgenesys: > that means the client must set up all-new credentials only through the Command Line tools.

Not necessarily. Client secrets are always 43 characters in length. If the obfuscation method will always generate a value that is not 43 characters, then users can put their secret directly in the file and the SDK can know that loading a secret from the config that is exactly 43 characters is a plain value. It can then use it as-is for requests and update the config file to store the obfuscated value for future use.

resources/sdk/clisdkclient/extensions/cmd/usage/usagequery_test.go
```diff
@@ -0,0 +1,46 @@
+package usage
```
ronanwatkins, line 1: Testing against swagger definition here. 
Currently only using path (including path params), method and description. 
I'll add tests for query params etc if they're ever required.

resources/sdk/clisdkclient/extensions/models/httperror.go
```diff
@@ -0,0 +1,15 @@
+package models
```
ronanwatkins, line 1: Moved this to fix import cycle error

resources/sdk/clisdkclient/extensions/config/config.go
carnellj: Yeah that sounds clunky.  I put my AWS credentials in the .cli command and they do not perform obfuscation.

timsmithgenesys (approved):

PR #98, opened 2021-01-27
Add 'module' field to package.json

Retsam:
The 'module' field is a newer alternative to 'jsnext:main', and helps allow the package to be consumed as raw ES Modules.  

We ran into this issue when trying to consume this package through Snowpack, which is a non-bundling build tool for JS development, and prefers to consume dependencies as ES Modules, rather than in AMD or CJS formats.


timsmithgenesys:
Thanks for adding this! I've changed the PR to merge into a branch in this repo so I can add a few additional changes to support this before merging into master. We'll get this deployed in a new version shortly.

PR #99, opened 2021-01-28
Add operation to command usage templates

ronanwatkins:
image


carnellj (changes requested):
Couple of quick things.

1.  Nice
2. Can you please move the operations piece to the very bottom of the help section.  I really want usage and flags first.
3.  Can you also add the permissions required to use this role right after operations.

carnellj (approved):

PR #100, opened 2021-01-28
Add /api/v2/authorization/roles and /api/v2/authorization/roles/{roleId} resources

ronanwatkins:


carnellj (changes requested):
Looks greats.  Thanks for picking this up.  I had planned on finishing this requirement today, but you beat me to it.
So can you also add the Update operation (Put for updating a role).

We also need to add and remove users from a role.

GET
/api/v2/authorization/roles/{roleId}/users
Get a list of the users in a specified role.


PUT
/api/v2/authorization/roles/{roleId}/users/add
Sets the users for the role
PUT
/api/v2/authorization/roles/{roleId}/users/remove
Removes the users from the role

Lets follow a similar pattern where users are there

carnellj (approved):

PR #106, opened 2021-02-02
Fix links to the wiki in the README

Retsam:
Fixes the README links to point to the current wiki, and adds a slightly more specific link under "Generating the SDK locally".


ronanwatkins (approved):
Nice work. Thanks!

timsmithgenesys (approved):

PR #114, opened 2021-02-10
Golang and Dotnet refresh token fix

ronanwatkins:


carnellj (approved):

resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -150,6 +150,9 @@ func (c *APIClient) CallAPI(path string, method string,
 		if err != nil {
 			return nil, err
 		}
+		if headerParams != nil {
```
timsmithgenesys, line 153: Why are these explicit assignments needed? Shouldn't the header be set above as a default header on line 94?

ronanwatkins: This is how the authorization header is set by the API call (line 78 of api.mustache). If you'd like to differ from that style and set it in the DefaultHeaders here I can make that change.

timsmithgenesys: Ah, I see. No worries. Some of the other SDKs apply auth via the default headers, but no need to change how this works since it's not already doing it.

timsmithgenesys (approved):

PR #128, opened 2021-02-19
Changing pagination behaviour for CLI

ronanwatkins:
Default behaviour is now to not paginate unless --autopaginate is set
Paginated results can be streamed with --stream
JSON schema of 200 response is used to determine if a resource is paginatable


carnellj (approved):
Hey Ronan,

Looks good.  Before we merge this, talk to Hakan.  We need to chat with Hakan and make sure that he modifies the test cases appropriately.  Also, before we release I need to update the recipes

PR #132, opened 2021-02-23
Prevent unmarshal of empty responses

joefruland:
Some APIs such as PostRoutingQueueMembers return an empty response on success, and attempting to unmarshal an empty body causes json.Unmarshal to return an error when there should not be one.


timsmithgenesys:
PostRoutingQueueMembers is annotated as returning a string response. If that's not what it returns, the annotations on the API resource need to be updated to reflect that it doesn't return anything and the SDK classes will be generated to not look for a response body. As a general rule, we don't implement workarounds in the SDK for API resources that are annotated incorrectly.

From swagger: 

image

PR #136, opened 2021-02-25
DEVENGAGE-489 Configurable tag format and changing Go SDK to be a go module

ronanwatkins:
- Allowing configurable git tag format and applying it to the Go SDK
- Changing Go SDK to be a go module

### DIFF
--- a/modules/builder.js
+++ b/modules/builder.js
@@ -507,11 +507,15 @@ function createRelease() {
 			githubApi.config.owner = repoOwner;
 			githubApi.config.token = getEnv('GITHUB_TOKEN');
 
+			const tagName = _this.config.settings.sdkRepo.tagFormat ?
+								_this.config.settings.sdkRepo.tagFormat.replace("{version}",  _this.version.displayFull) :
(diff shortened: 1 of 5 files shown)


carnellj (approved):

carnellj (approved):

modules/builder.js
```diff
@@ -507,11 +507,15 @@ function createRelease() {
 			githubApi.config.owner = repoOwner;
 			githubApi.config.token = getEnv('GITHUB_TOKEN');
 
+			const tagName = _this.config.settings.sdkRepo.tagFormat ?
```
timsmithgenesys, line 510: Use `maybeInit` in the constructor to set a default value of `{version}` so you don't have to have logic here. Other defaults: https://github.com/purecloudlabs/platform-client-sdk-common/blob/master/modules/builder.js#L72

This is just an optional suggestion, but will help reduce validation and conditional setting logic needed in the code.

timsmithgenesys (approved):

PR #142, opened 2021-03-05
DEVENGAGE-501-Fix-newVersion-variable

WitLi07:
newVersion variable should be false when the building branch is not master.

The only functional change is at Line 282. Others are prettier auto-format changes.

### DIFF
--- a/modules/builder.js
+++ b/modules/builder.js
@@ -16,7 +16,6 @@ const swaggerDiff = require('./swaggerDiff');
 const git = require('./gitModule');
 const zip = require('./zip');
 
-
 /* PRIVATE VARS */
 
 let _this;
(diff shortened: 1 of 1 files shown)


modules/builder.js
```diff
@@ -286,12 +279,10 @@ function prebuildImpl() {
 				var newVersion = swaggerDiff.stringifyVersion(_this.version, true);
 
 				// Determine if new version
-				_this.isNewVersion = oldVersion !== newVersion;
+				_this.isNewVersion = getEnv('BRANCH_NAME') !== 'master' ? false : oldVersion !== newVersion;
```
WitLi07, line 282: The only functional change is here.

ronanwatkins (approved):

timsmithgenesys (approved):

PR #145, opened 2021-03-12
Devengage 527 add some detail http info

carnellj:


timsmithgenesys (commented):
This should be added to the readmes for the other SDKs as well. At least the java-guest and web messaging SDKs will share this exact behavior. Other SDKs have some similar functionality that may be worth pointing out..

resources/sdk/purecloudjava/templates/README.mustache
```diff
@@ -258,6 +258,12 @@ System.out.println(meWithHttpInfo.getHeaders());
 System.out.println(meWithHttpInfo.getCorrelationId());
 System.out.println(meWithHttpInfo.getBody().getName());
 ```
+:::{"alert":"warning","title":"Using the Extended HTTP APIs and exception handling"}
+The default behavior for the our Standard Genesys Cloud Java SDK is that they will automatically throw an exception for 429 and 5xx level errors.
```
ronanwatkins, line 262: for the our -> for the
that they will -> that it will

PR #147, opened 2021-03-15
Remove default header trace in Go SDK

joefruland:
Prevent always tracing default headers as this spams logs when one is added. If needed, they will be traced on line 133 along with the full Request when the debug setting is enabled.


resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -93,7 +93,6 @@ func (c *APIClient) CallAPI(path string, method string,
 	// Set default headers
 	if c.configuration.DefaultHeader != nil {
 		for k, v := range c.configuration.DefaultHeader {
-			fmt.Printf("  %v=%v", k, v)
```
timsmithgenesys, line 96: Thanks, this shouldn't be in here. The intent for the SDKs to be silent under normal circumstances.

timsmithgenesys (approved):

PR #148, opened 2021-03-16
DEVENGAGE-490-responsemgt-fail-without-libraryId

WitLi07:
responsemgt issue fixed. Now when invoking `gc responsemgt responses list` without providing libraryId, it will bring up help page and asking users to provide the flag. (example below)
image


resources/sdk/clisdkclient/templates/api.mustache
```diff
@@ -38,7 +39,7 @@ func Cmd{{classVarName}}() *cobra.Command { {{#operation}}
 
 {{#operation}}
 var {{operationId}}Cmd = &cobra.Command{
-	Use:   "{{operationId}}{{#pathParams}} [{{paramName}}{{#hasMore}}{{/hasMore}}]{{/pathParams}}",
+	Use:   "{{operationId}}{{#pathParams}} [{{paramName}}{{#hasMore}}{{/hasMore}}]{{/pathParams}}{{#queryParams}}{{#required}} [{{paramName}}]{{/required}}{{/queryParams}}",
```
ronanwatkins, line 42: Probably best to leave this change out. This would confuse people into thinking it was a path param.
Marking the flag as required would be enough to let people know to add it

ronanwatkins (changes requested):

resources/sdk/clisdkclient/templates/api.mustache
WitLi07: Yeah I also thought about that and not entirely sure if we should have the required query parameter there. But for the usage, we use curly brackets for path params and square brackets for query params. And since it's telling users how to use the command, maybe we should include the required parameters, but using different syntax to represent query params? I'm okay with either one. Any thoughts on this? @carnellj

carnellj: Agree with Ronan.  This is an edge case.  Very few of the commands have required queryParameters.  This is probably a situation where someone forgot to add it to the body and quick added it as a queryParam

WitLi07: Sure, I'll remove that query parameter part then.

WitLi07: Fixed. Removed the query param section in usage.

ronanwatkins (approved):

carnellj (approved):

PR #149, opened 2021-03-17
Remove ispreview in README

WitLi07:


timsmithgenesys (approved):

resources/sdk/clisdkclient/templates/README.mustache
```diff
@@ -1,6 +1,6 @@
 ---
 title: Platform API CLI SDK
-ispreview: true
+ispreview: false
```
timsmithgenesys, line 3: FWIW, this doesn't need to be specified at all; false and not specifying the parameter at all are equivalent. The code only checks to see if a truthy value has been set, otherwise the preview banner is not shown.

WitLi07: Removed

PR #150, opened 2021-03-18
DEVENGAGE-561 Several fixes for Go SDK

ronanwatkins:
Fix for []string query params
Fix for unicode escaping in deserialized models
Fix for broken README link (also in other SDKs)
Fix for go.mod and go.sum location


timsmithgenesys (approved):

resources/sdk/purecloudgo/templates/api.mustache
```diff
@@ -81,29 +81,8 @@ func (a {{classname}}) {{nickname}}({{#allParams}}{{paramName}} {{{dataType}}}{{
...
-			queryParams["{{paramName}}"] = value
-		}
-	} else {
-		queryParams["{{paramName}}"] = a.Configuration.APIClient.ParameterToString({{paramName}}, collectionFormat)
-	}
-	{{/isListContainer}}
-	{{^isListContainer}}
-	collectionFormat = ""
-	if str, ok := interface{}({{paramName}}).(string); ok {
-		if str != "" {
```
joefruland, line 100: I think removing this empty string check for non-list params will break API calls that are relying on it to avoid adding optional query parameters. This should remain until the optional parameter changes to switch to pointers are in.

joefruland: Nevermind, it looks like this logic was moved to apiclient line 85.

PR #153, opened 2021-03-23
DEVENGAGE-472 - Adding file config and logging to the dotnet SDK

ronanwatkins:


resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -248,6 +249,9 @@ namespace {{packageName}}.Client
                 }
             }
 
+            if ((int)response.StatusCode < 200 || (int)response.StatusCode >= 300)
```
carnellj, line 255: Why do you have to do the cast here.  Doesnt .NET do autoboxing of values.  If not I would be surprised if their Integer object does not have a toInt() method. on it.

resources/sdk/pureclouddotnet/templates/Configuration.mustache
```diff
@@ -62,6 +66,47 @@ namespace {{packageName}}.Client
...
+            if (AutoReloadConfig)
+            {
+                ThreadStart childref = new ThreadStart(RunConfigChecker);
+                Console.WriteLine("In Main: Creating the Child thread");
+                Thread childThread = new Thread(childref);
+                childThread.Start();
+            }
+        }
+
+        private void applyConfigFromFile()
```
carnellj, line 120: Why is this function empty?

resources/sdk/pureclouddotnet/templates/Configuration.mustache
```diff
@@ -62,6 +66,47 @@ namespace {{packageName}}.Client
...
+        {
+
+        }
+
+        public static void RunConfigChecker()
+        {
+            while (true)
+            {
+                Console.WriteLine("runConfigChecker");
+                Thread.Sleep(2000);
```
carnellj: Just to double check...... I am assuming this thread. Sleep was running off the thread you spun off for checking the config.  Also, don't hard code the value using a magic number.  Put 2000 in a constant at the top of the file so we know exactly why this parameter is there.

resources/sdk/pureclouddotnet/templates/Logger.mustache
```diff
@@ -0,0 +1,312 @@
...
+                    }
+                }
+            }
+        }
+
+        private static string requestBodyToString(object requestBody)
+        {
+            if (requestBody != null)
+            {
+            //     if (requestBody.GetType() == typeof(String))
```
carnellj: Why is this code commented out?  Remove uncommented code unless you are planning to use it.

resources/sdk/pureclouddotnet/templates/Logger.mustache
```diff
@@ -0,0 +1,312 @@
...
+
+        public int StatusCode { get; set; }
+
+        public string RequestBody { get; set; }
+
+        public string ResponseBody { get; set; }
+
+        public string AsString(LogFormat logFormat, bool logRequestBody, bool logResponseBody)
+        {
+            RequestHeaders["Authorization"] = "[REDACTED]";
```
carnellj: Nice

carnellj (approved):
Couple of small changes.  Please fix or let me know why they need to stay

resources/sdk/pureclouddotnet/templates/Configuration.mustache
ronanwatkins: I think you're viewing the version from the first commit. If you look at the changes from https://github.com/purecloudlabs/platform-client-sdk-common/pull/153/files it should be all up-to-date

WitLi07 (approved):

jacob-shaw-genesys (approved):

PR #154, opened 2021-03-25
DEVENGAGE-472 - Adding file config and logging to the golang SDK

ronanwatkins:


resources/sdk/purecloudgo/templates/api.mustache
```diff
@@ -17,16 +17,15 @@ type {{classname}} struct {
 
 // New{{classname}} creates an API instance using the default configuration
 func New{{classname}}() *{{classname}} {
+	fmt.Sprintf(strings.Title(""), "")
```
ronanwatkins, line 20: Putting this redundant function call here to make use of the "fmt" and "strings" packages. They are only used for certain API calls and will cause a compiler error if they are imported and not used

resources/sdk/purecloudgo/extensions/config.toml
```diff
@@ -0,0 +1,6 @@
+# Just leaving this here for now
+# Logging
+log_level="trace"
+log_format="text"
+log_to_console="true"
+log_file_path="/Users/ronanwatkins/genesys_src/repos/platform-client-sdk-common/output/purecloudgo/platformclientv2/log.log"
```
carnellj: Might want to change this to a Generic value rather then hardcoded to your path,

carnellj (approved):

resources/sdk/purecloudgo/extensions/config.toml
ronanwatkins: That's from an earlier commit. All the fresh code is here: https://github.com/purecloudlabs/platform-client-sdk-common/pull/154/files

resources/sdk/purecloudgo/templates/logger.mustache
```diff
@@ -0,0 +1,248 @@
...
+	logStatement := &logStatement{
+		Date:            &now,
+		Level:           "trace",
+		Method:          method,
+		URL:             URL,
+		RequestBody:     string(requestBody),
+		StatusCode:      statusCode,
+		RequestHeaders:  requestHeaders,
+		ResponseHeaders: responseHeaders,
+	}
```
WitLi07, line 170: According to the design doc, I think trace level log should also contain a correlation Id?

ronanwatkins: You're right, thanks for pointing that out

PR #156, opened 2021-03-25
Expanding list of retryable methods in the python SDK

ronanwatkins:


resources/sdk/purecloudpython/templates/rest.mustache
```diff
@@ -102,12 +102,15 @@ class RESTClientObject(object):
         proxy_username = Configuration().proxy_username
         proxy_password = Configuration().proxy_password
 
+        retries = urllib3.util.Retry()
+        retries.method_whitelist = {'DELETE', 'GET', 'HEAD', 'OPTIONS', 'POST', 'PUT', 'PATCH', 'TRACE'}
```
timsmithgenesys, line 106: I'm don't think we need HEAD, OPTIONS, or TRACE. TRACE and HEAD aren't used for API resources. Neither is OPTIONS, which is only used for CORS requests which aren't valid for python since it's not a web browser. It's fine to leave them in if there's any case where they may be used, but I can't think of one.

timsmithgenesys (approved):

resources/sdk/purecloudpython/templates/rest.mustache
carnellj: Also, I am not sure about behavior.  If we leave them out I believe they wont retry on these.

carnellj (commented):
Looks good

PR #157, opened 2021-03-25
DEVENGAGE-472 - Adding file config and logging to the python SDK

ronanwatkins:


resources/sdk/purecloudpython/templates/README.mustache
```diff
@@ -111,6 +111,82 @@ PureCloudPlatformClientV2.configuration.proxy_password = 'YOUR_PROXY_PASSWORD'
 
 The Python SDK uses `urllib3.ProxyManager` to make requests when `proxy` is given.
 
+#### SDK Logging
+
+Logging of API requests and responses can be controlled by a number of parameters on the `configuration`'s `logging` instance.
+
+`PureCloudPlatformClientV2.logger.LogLevel` values:
+1. LTrace (HTTP Method, URL, Request Body, HTTP Status Code, Request Headers, Response Headers)
```
carnellj, line 119: Why are we putting an L in front of the method names?  Will this be consistent across all of the SDKs in terms of method names.

resources/sdk/purecloudpython/templates/README.mustache
```diff
@@ -111,6 +111,82 @@ PureCloudPlatformClientV2.configuration.proxy_password = 'YOUR_PROXY_PASSWORD'
...
+        "refresh_access_token": true,
+        "refresh_token_wait_max": 10
+    },
+    "general": {
+        "live_reload_config": true,
+        "host": "https://api.mypurecloud.com"
+    }
+}
+```
+
```
carnellj, line 189: Do we need a mechanism to show how the trace levels would be set programmatically via the config object.  I always thought the file was the first option and then programmatic config is the second option.

carnellj (changes requested):

resources/sdk/purecloudpython/templates/README.mustache
ronanwatkins: Yeah, this code block shows how all the logging config can be set programmatically: https://github.com/purecloudlabs/platform-client-sdk-common/pull/157/files#diff-58afaa4a115cb1ef50340943bb96aa5fc4e2907723a2ecda2bb71d1c7d4eab86R133

resources/sdk/purecloudpython/templates/README.mustache
ronanwatkins: Just in front on enum names. I've applied that consistently across all SDKs

WitLi07 (approved):

resources/sdk/purecloudpython/templates/README.mustache
```diff
@@ -111,6 +111,82 @@ PureCloudPlatformClientV2.configuration.proxy_password = 'YOUR_PROXY_PASSWORD'
...
+`PureCloudPlatformClientV2.logger.LogFormat` values:
+1. JSON
+2. TEXT - default
+
+By default, the request and response bodies are not logged because these can contain PII. Be mindful of this data if choosing to log it.  
+To log to a file, provide a file path to the the `PureCloudPlatformClientV2.configuration.logger.log_file_path` property. SDK users are responsible for the rotation of the log file.
+
+Example logging configuration:
+```python
+PureCloudPlatformClientV2.configuration.logger.log_level = PureCloudPlatformClientV2.logger.LogLevel.LError
```
WitLi07, line 133: On the right side, it should be `PureCloudPlatformClientV2.configuration.logger.LogLevel.LError`

resources/sdk/purecloudpython/templates/README.mustache
```diff
@@ -111,6 +111,82 @@ PureCloudPlatformClientV2.configuration.proxy_password = 'YOUR_PROXY_PASSWORD'
...
+PureCloudPlatformClientV2.configuration.logger.log_request_body = True
+PureCloudPlatformClientV2.configuration.logger.log_response_body = True
+PureCloudPlatformClientV2.configuration.logger.log_format = PureCloudPlatformClientV2.logger.LogFormat.TEXT
```
WitLi07, line 136: Same as line 133, should be `configuration.logger` on the right side

resources/sdk/purecloudpython/templates/README.mustache
ronanwatkins: I've just double checked this; it works like this because `configuration` holds an instance of the logger but to access the actual logger class you need to use `PureCloudPlatformClientV2.logger`.

PR #174, opened 2021-05-03
Use SetBody on retryablerequest to ensure body can be resent

joefruland:
Using the SetBody method is required to ensure the the retryablerequest library is able to rewind the body for retries. Otherwise it sends an empty body in the first retry. https://github.com/hashicorp/go-retryablehttp/blob/v0.7.0/client.go#L126


ronanwatkins (approved):
Hadn't known about that. Thanks for the help.

PR #182, opened 2021-05-12
DEVENGAGE-571-javascript-logger

WitLi07:
Fix JS SDK browser not working issue


resources/sdk/purecloudjavascript/templates/configuration.mustache
```diff
@@ -0,0 +1,168 @@
...
+
+	setConfigPath(path) {
+		if (path && path !== this.configPath) {
+			this.configPath = path;
+			this.liveLoadConfig();
+		}
+	}
+
+	updateConfigFromFile() {
+		const ConfigParser = require('configparser');
```
WitLi07, line 72: This line should be the only change compared to the previous PR. This line used to be at the top of the file, moved it here for browser compatibility.

ronanwatkins: Is that all that needed to be fixed :D

ronanwatkins (approved):

PR #185, opened 2021-05-18
Actually fixing the Go SDK build

ronanwatkins:


resources/sdk/purecloudgo/scripts/postbuild-postrun.js
```diff
@@ -8,6 +8,14 @@ try {
 	console.log('repoPath', repoPath);
 	console.log('buildDir', buildDir);
 
+	const dir = fs.opendirSync(repoPath);
```
ronanwatkins, line 11: This bit is needed to delete directories so old files aren't left in them

WitLi07 (approved):

PR #190, opened 2021-05-24
Devengage 644 java sdk snyk issues

jacob-shaw-genesys:


timsmithgenesys (approved):

resources/sdk/purecloudjava-guest/extensions/connector/ning/AsyncHttpClientConnectorProvider.java
```diff
@@ -39,7 +42,7 @@ public ApiClientConnector create(ApiClientConnectorProperties properties) {
                 @Override
                 public void connectFailed(URI uri, SocketAddress sa, IOException ioe) { }
             };
-            ProxyServerSelector proxyServerSelector = ProxyUtils.createProxyServerSelector(proxySelector);
+            ProxyServerSelector proxyServerSelector = createProxyServerSelector(proxySelector);
```
ronanwatkins, line 45: Looks like you should be using [public static ProxyServerSelector createProxyServerSelector(Properties properties)](https://github.com/AsyncHttpClient/async-http-client/blob/9b7298b8f1cb41fed5fb5a1315267be323c875d6/client/src/main/java/org/asynchttpclient/util/ProxyUtils.java#L100). You could probably cast your ApiClientConnectorProperties into java.util.Properties for the method call, or copy them into an instance of that type if that doesn't work

ronanwatkins: If there's a reason you can't use that then this solution is grand.

ronanwatkins (changes requested):

carnellj (approved):

resources/sdk/purecloudjava-guest/extensions/connector/ning/AsyncHttpClientConnectorProvider.java
diemenschmaschine92: I was able to get it working this way.  My local client still runs successfully using the static method.

ronanwatkins (approved):

PR #198, opened 2021-06-03
Serialize error response details as an array of objects

joefruland:
The Go SDK is returning JSON serialization errors when details are included in an error response.

### DIFF
--- a/resources/sdk/purecloudgo/templates/apiresponse.mustache
+++ b/resources/sdk/purecloudgo/templates/apiresponse.mustache
@@ -40,6 +40,13 @@ func (r *APIResponse) SetError(err *APIError) {
 	r.ErrorMessage = fmt.Sprintf("API Error: %v - %v (%v)", r.StatusCode, err.Message, r.CorrelationID)
 }
 
+type APIErrorDetail struct {
+	ErrorCode  string `json:"errorCode,omitempty"`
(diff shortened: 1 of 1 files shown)


resources/sdk/purecloudgo/templates/apiresponse.mustache
```diff
@@ -48,7 +55,7 @@ type APIError struct {
 	MessageParams     map[string]interface{} `json:"messageParams,omitempty"`
 	Code              string                 `json:"code,omitempty"`
 	ContextID         string                 `json:"contextId,omitempty"`
-	Details           []string               `json:"details,omitempty"`
+	Details           []APIErrorDetail       `json:"details,omitempty"`
```
ronanwatkins, line 58: The APIErrorDetail struct is defined as `Detail` in https://github.com/MyPureCloud/platform-client-sdk-go/blob/master/platformclientv2/detail.go.
You could update this to make use of that struct, either way works.

ronanwatkins (approved):

resources/sdk/purecloudgo/templates/apiresponse.mustache
joefruland: Ah I missed that. I'll switch to the existing struct, thanks.

PR #200, opened 2021-06-16
Go SDK: Do not unmarshal string responses

ronanwatkins:


resources/sdk/purecloudgo/templates/api.mustache
```diff
@@ -121,7 +121,11 @@ func (a {{classname}}) {{nickname}}({{#allParams}}{{paramName}} {{{dataType}}}{{
 	} else if err == nil && response.Error != nil {
 		err = errors.New(response.ErrorMessage)
 	}{{#returnType}} else {
-		err = json.Unmarshal([]byte(response.RawBody), &successPayload)
+		if "{{{returnType}}}" == "string" {
```
timsmithgenesys, line 124: A couple additional cases to consider:

1. Does Unmarshal handle root-level arrays?
2. Will either of these fail if the response is empty (successPayload is nil)? There are resources that document returning a string but don't actually send a value. I think most of those were cleaned up, but make sure a nil value won't blow up here.

timsmithgenesys (approved):

resources/sdk/purecloudgo/templates/api.mustache
ronanwatkins: 1. Yeah, no problem there.
2. Would that present itself as response.RawBody being empty or "{}"? Both of those cases would be fine

timsmithgenesys: Sounds good, thanks for confirming!

PR #208, opened 2021-07-13
Update SDK

dasan03:


resources/sdk/webmessagingjava/config.json
```diff
@@ -14,7 +14,7 @@
 		"swagger": {
 			"internalSwaggerPath": "https://api.inindca.com/api/v2/docs/swaggerall",
 			"oldSwaggerPath": "${SDK_REPO}/swagger.json",
-			"newSwaggerPath": "https://s3.amazonaws.com/inin-prod-use1-purecloud-shared/webmessaging/swagger/swagger-latest.json",
+			"newSwaggerPath": "https://s3.amazonaws.com/inin-dev-use1-purecloud-shared/webmessaging/swagger/swagger-latest.json",
```
timsmithgenesys, line 17: Public SDKs must be built against production. You can use an alternate swagger definition for the internal SDK in the SDK config repo in bitbucket.

timsmithgenesys (changes requested):

resources/sdk/webmessagingjava/config.json
dasan03: on which repo I should include DCA ? there are some definition not yet in production

PR #214, opened 2021-07-21
Adding experimental features en/disablement from the cli

charliecon:
Adding the cli command which allows user to change a boolean called experimental_features_enabled inside config.toml


.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
timsmithgenesys, line 1: Don't check in anything in the .idea folder. Fix this by commiting the removal of the folder. Then in a separate commit, add `.idea/` to the .gitignore file.

timsmithgenesys (commented):
What does this feature actually do? It doesn't seem like it has an effect on anything other than managing the setting.

ronanwatkins:
> What does this feature actually do? It doesn't seem like it has an effect on anything other than managing the setting.

Yeah, there are no experimental features yet. I was thinking it's probably best to hold off merging until there's an experimental feature we want to add.

carnellj:
> > What does this feature actually do? It doesn't seem like it has an effect on anything other than managing the setting.
> 
> Yeah, there are no experimental features yet. I was thinking it's probably best to hold off merging until there's an experimental feature we want to add.

I disagree with holding off.  Add a dummy feature and don't document it.  Then we can test turning on and off the feature.  We are going to be using this feature soon enough.  Besides a dummy feature, we can have Hakan add some dummy tests.

resources/sdk/clisdkclient/extensions/cmd/root_test.go
```diff
@@ -0,0 +1,29 @@
+package cmd
+
+import "testing"
+
+func TestGetProfileName(t *testing.T) {
+	var tests = []struct {
```
ronanwatkins, line 6: That's pretty cool :)

ronanwatkins (approved):

resources/sdk/clisdkclient/extensions/cmd/experimental/experimental.go
```diff
@@ -0,0 +1,70 @@
...
+		profileName, _ := cmd.Root().Flags().GetString("profile")
+		experimental_features.ListAllFeatures(profileName)
+	},
+}
+
+func setFeature(cmd *cobra.Command, featureCommand string, enabled bool) {
+	profileName, _ := cmd.Root().Flags().GetString("profile")
+
+	if !experimental_features.IsValidCommand(featureCommand) {
+		fmt.Printf("Command '%v' does not exists. All experimental features are listed below:\n", featureCommand)
```
ronanwatkins, line 61: does not exists -> does not exist

resources/sdk/clisdkclient/extensions/config/config.go
```diff
@@ -224,6 +224,19 @@ func SetLoggingEnabled(c Configuration, loggingEnabled bool) error {
...
 
+func SetExperimentalFeature(profileName string, featureName string, enabled bool) error {
+	viper.Set(fmt.Sprintf("%s.%s_enabled", profileName, featureName), enabled)
+	return viper.WriteConfig()
+}
+
+func GetExperimentalFeature(profileName string, featureName string) bool {
+	err := viper.ReadInConfig()
+	if err != nil {
+		fmt.Println(err)
```
ronanwatkins, line 235: One last suggestion, you could use logger.Error() to print this error. Also good to explicitly return false here.

charliecon: Do you mean I could import logger.go? I only see a Fatal method there. Or do you mean fmt.Errorf?

ronanwatkins: Sorry, there's no error method. logger.Warn would be good here. It would mean that the error wouldn't be printed to the console if this function returns an error.
This function will return an error if no config file exists, which will usually be the case when people use the CLI in pipelines and we don't want this error message confusing things.

PR #218, opened 2021-08-03
adding ability to send and receive data as yaml

michaeljroddy:


resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -3,16 +3,12 @@ package cmd
 import (
 	"github.com/mypurecloud/platform-client-sdk-cli/build/gc/config"
 	"github.com/mypurecloud/platform-client-sdk-cli/build/gc/logger"
-	"github.com/mypurecloud/platform-client-sdk-cli/build/gc/models"
```
ronanwatkins, line 6: Looks like you've removed some necessary stuff from this file.
You can build with `node sdkBuilder.js --sdk clisdkclient` to make sure everything is ok before pushing your changes

resources/sdk/purecloudpython/templates/api_client.mustache
```diff
@@ -726,16 +726,24 @@ class ApiClient(object):
         :return: date.
         """
         try:
-            from dateutil.parser import parse
```
ronanwatkins, line 729: Not sure how this got added. Best to undo changes to this file.

resources/sdk/clisdkclient/extensions/utils/general.go
```diff
@@ -305,3 +308,14 @@ func MilliSecondsToNanoSeconds(milliSeconds int64) time.Duration {
 func SecondsToNanoSeconds(seconds int) time.Duration {
 	return MilliSecondsToNanoSeconds(int64(seconds)) * 1000
 }
+
+func ConvertToJSON(data string) string {
+	if data_format.InputFormat == "json" {
```
ronanwatkins, line 313: Would be nice to do a case insensitive comparison to be more forgiving to the user. https://stackoverflow.com/a/67818307 would be a nice way of comparing any time you're checking for "json" or "yaml" with this feature

resources/sdk/clisdkclient/extensions/utils/render.go
```diff
@@ -1,12 +1,30 @@
...
 
+	"github.com/mypurecloud/platform-client-sdk-cli/build/gc/data_format"
 	"github.com/tidwall/pretty"
+	"sigs.k8s.io/yaml"
 )
 
 func Render(data string) {
-	result := pretty.Pretty([]byte(data))
-	fmt.Printf("%s", result)
+	if data_format.OutputFormat == "json" && IsJSON(data) {
```
ronanwatkins, line 13: Looks like nothing will be printed if the data isn't JSON. You could leave it printing the data and returning if the data isn't JSON.
Then you just need an if/else checking if the OutputFormat is json or yaml and render as necessary

resources/sdk/clisdkclient/extensions/data_format/data_format.go
```diff
@@ -0,0 +1,4 @@
+package data_format
+
+var InputFormat string
```
ronanwatkins, line 3: Multiple variables can be defined using a block like so:
```
var (
)
```

PR #219, opened 2021-08-05
Added logic to check if new version exists inside gc version command.

charliecon:


resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -12,16 +12,29 @@ import (
 	"flag"
 	"fmt"
 	"strings"
+	"encoding/json"
+    "io/ioutil"
```
ronanwatkins, line 16: There's a formatting issue here. Also in other parts of the file. Have a look at some of the other code added below.
The issue is probably due to a mix of tabs and spaces. It can look fine from VSCode but the spacing is different here so it shows up.

resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -12,16 +12,29 @@ import (
...
+type LatestVersion struct {
+	Version string `json:"version"`
+}
+
 var (
 	profileName string
 	output string
 )
 
+var (
```
ronanwatkins, line 33: These variable declarations can be declared in the var block above. Usually the style is to have a single var block per file

resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -35,9 +48,52 @@ var versionCmd = &cobra.Command{
...
 		fmt.Println("{{packageVersion}}")
+
+        var latestVersion string
+        if strings.Contains("{{basePath}}", "api.mypurecloud.com") {
+            latestVersion = retrieveLatestVersion(externalVersionURL)
+        } else {
+            latestVersion = retrieveLatestVersion(internalVersionURL)
+        }
+
+        if latestVersion == "{{packageVersion}}" {
```
ronanwatkins, line 59: You'll have to write a function here to compare semantic version numbers. It would be a good idea to add a unit test to root_test to test the function.

resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -35,9 +48,52 @@ var versionCmd = &cobra.Command{
 	Long:  `All software has versions. This is gc version's`,
 	Run: func(cmd *cobra.Command, args []string) {
 		fmt.Println("{{packageVersion}}")
```
ronanwatkins, line 50: Might be nice to change this to "Current version: {{packageVersion}}" now that there's going to be more printed out below

resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -35,9 +48,52 @@ var versionCmd = &cobra.Command{
...
 	},
 }
 
+func retrieveLatestVersion(url string) string {
+	client := http.Client{
+		Timeout: time.Second * 2,
+	}
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		logger.Fatal(err)
```
ronanwatkins, line 73: It's probably best to have retrieveLatestVersion return an error and a string and return the error and an empty string if an error is encountered in the function.
Then have the caller check if the error isn't nil and print something like "error retrieving latest version".

resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -34,10 +44,68 @@ var versionCmd = &cobra.Command{
...
+		return jsonErr, ""
+	}
+
+	return nil, latestVersion.Version
+}
+
+func versionsAreEqual(current string, latest string) bool {
+	v1, _ := version.NewVersion(current)
+	v2, _ := version.NewVersion(latest)
+	return v1.Equal(v2)
```
ronanwatkins, line 106: GreaterThanOrEqual would be safest here in case for some reason the latest version file isn't updated with the latest version or there's a caching issue.

Also it would be a good idea to write a unit test for this function.

resources/sdk/clisdkclient/templates/root.mustache
```diff
@@ -34,10 +44,68 @@ var versionCmd = &cobra.Command{
...
+	if versionsAreEqual("17.0.0", latestVersion) {
+		fmt.Println("You're all up to date.")
+	} else {
+		fmt.Printf("A new version of the CLI is available: %v\n", latestVersion)
+	}
+}
+
+func retrieveLatestVersion(url string) (error, string) {
+	client := http.Client{
+		Timeout: time.Second * 2,
```
ronanwatkins, line 75: The timeout could go up a small bit.

PR #220, opened 2021-08-05
New send and receive data as yaml

michaeljroddy:


resources/sdk/clisdkclient/extensions/experimental_features/experimental_features.go
```diff
@@ -20,6 +21,9 @@ func IsValidCommand(command string) bool {
 	if command == "dummy_command" {
 		return true
 	}
+	if command == "alternative_formats" {
```
ronanwatkins, line 24: You can remove this bit. The check above is only necessary for the dummy_command

resources/sdk/clisdkclient/extensions/models/experimentalfeaturecommands.go
```diff
@@ -2,8 +2,9 @@ package models
```
ronanwatkins, line 2: You can go ahead and do some refactoring here. `allCommands` declared in resources/sdk/clisdkclient/extensions/experimental_features/experimental_features.go should be declared here as a global variable. It should just hold AlternativeFormats at the moment.

resources/sdk/clisdkclient/extensions/utils/general.go
```diff
@@ -305,3 +309,19 @@ func MilliSecondsToNanoSeconds(milliSeconds int64) time.Duration {
 func SecondsToNanoSeconds(seconds int) time.Duration {
 	return MilliSecondsToNanoSeconds(int64(seconds)) * 1000
 }
+
+func convertToJSON(data string) string {
+	if strings.EqualFold("json", data_format.InputFormat) && isJSON(data) {
+		return data
+	}
+	result, err := yaml.YAMLToJSON([]byte(data))
```
ronanwatkins, line 317: It would be safest to only convert yaml to json if the inputformat flag specifies that

resources/sdk/clisdkclient/extensions/utils/general.go
```diff
@@ -305,3 +309,19 @@ func MilliSecondsToNanoSeconds(milliSeconds int64) time.Duration {
...
+	if err != nil {
+		logger.Fatal(fmt.Printf("err: %v\n", err))
```
ronanwatkins, line 319: Logging the result of fmt.Printf() would result in some weird output. Best to do something like "logger.Fatalf("Error converting YAML to JSON: %v\n", err)"

ronanwatkins (approved):
Works really well!

PR #221, opened 2021-08-17
Change date and datetime handling in the Go SDK

ronanwatkins:
Required to facilitate handling for 3 different date formats:
* yyyy-MM-ddTHH:mm:ss.SSS
* yyyy-MM-ddTHH:mm:ss.SSSZ
* yyyy-MM-dd

Related: https://github.com/purecloudlabs/swagger-codegen/pull/39

### DIFF
--- a/resources/sdk/purecloudgo/templates/model.mustache
+++ b/resources/sdk/purecloudgo/templates/model.mustache
@@ -2,6 +2,7 @@ package {{packageName}}
 {{#models}}
 import ({{#imports}}
 	"{{import}}"{{/imports}}
+	"github.com/leekchan/timeutil"
 	"encoding/json"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


resources/sdk/purecloudgo/templates/model.mustache
```diff
@@ -15,6 +16,45 @@ type {{classname}} struct { {{#vars}}
...
 
+func (u *{{classname}}) MarshalJSON() ([]byte, error) {
+	// Redundant initialization to avoid unused import errors for models with no Time values
+	_  = timeutil.Timedelta{}
+	type Alias {{classname}}
+
+	{{#vars}}{{#isDateTime}}
+	{{name}} := new(string)
+	if u.{{name}} != nil {
+		{{#vendorExtensions.x-local-date-time}}*{{name}} = timeutil.Strftime(u.{{name}}, "%Y-%m-%dT%H:%M:%S.%f"){{/vendorExtensions.x-local-date-time}}
```
timsmithgenesys, line 27: Are these valid if there's no milliseconds? There are 4 known formats used by the Java SDK: https://github.com/purecloudlabs/platform-client-sdk-common/blob/master/resources/sdk/purecloudjava/extensions/ApiDateFormat.java#L14.

timsmithgenesys: Any of them. I believe it's an API-layer behavior that truncates the milliseconds when they're exactly zero.

ronanwatkins: That should be ok then. This change was more for formatting times to be sent to the API as a request payload.
I doubt there would be any issues arising from a client expecting `2021-08-04T08:00:00Z`  and receiving `2021-08-04T08:00:00.00Z` instead.

timsmithgenesys: AFAIK that's accurate. I believe the API is tolerant of optional ms.

PR #230, opened 2021-09-08
Adding Authorization Code Grant to the cli.

charliecon:


resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -243,6 +269,48 @@ func authorize(c config.Configuration) (models.OAuthTokenData, error) {
...
+	}
+
+	http.HandleFunc(path, getAuthCode)
+	log.Fatal(http.ListenAndServe(port, nil))
+}
+
+func openBrowserForLogin(wg *sync.WaitGroup, clientID string, redirectURI string) {
+	encodedURI := strings.Replace(redirectURI, ":", "%3A", -1)
+	encodedURI = strings.Replace(redirectURI, "/", "%2F", -1)
+	loginURL := fmt.Sprintf("https://login.mypurecloud.com/oauth/authorize?client_id=%v&response_type=code&redirect_uri=%v", clientID, encodedURI)
```
ronanwatkins, line 294: The login URL wouldn't be `mypurecloud.com` for everyone. You can use the `Environment()` from the config to replace it

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -243,6 +269,48 @@ func authorize(c config.Configuration) (models.OAuthTokenData, error) {
...
+			return
+		}
+	}
+
+	http.HandleFunc(path, getAuthCode)
+	log.Fatal(http.ListenAndServe(port, nil))
+}
+
+func openBrowserForLogin(wg *sync.WaitGroup, clientID string, redirectURI string) {
+	encodedURI := strings.Replace(redirectURI, ":", "%3A", -1)
```
ronanwatkins, line 292: Pretty sure the "proper way" would be to generate the loginUrl with an unedited redirectURI and then use `url.Values.Encode()` to encode it. Example: https://golang.cafe/blog/how-to-url-encode-string-in-golang-example.html

PR #231, opened 2021-09-09
allowing user to specify an access token during create profile, or as an environment variable or as a parameter override

michaeljroddy:
1. Create profile flow updated to allow for either an access token, client credentials or both during create profile. If a user provides an access token, this will be used over client credentials.
2.  An environment variable GENESYS_CLOUD_ACCESS_TOKEN can also be used to set the token. This will take priority over an access token in config.
3. A parameter override --accesstoken can also be used to set the token. This will take priority over a token set as an environment variable or a token in config.

### DIFF
--- a/resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
+++ b/resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
@@ -74,7 +74,7 @@ func requestUserInput() config.Configuration {
 	}
 
 	fmt.Printf("Please provide either an Access Token, Client Credentials (Client ID and Client Secret) or both\n")
-	fmt.Printf("Note: If you provide an Access Token, this will be used over Client Credentials")
+	fmt.Printf("Note: If you provide an Access Token, this will be used over Client Credentials\n")
 	for {
 		fmt.Printf("Access Token: ")
 		fmt.Scanln(&accessToken)
--- a/resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
+++ b/resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
@@ -2,18 +2,19 @@ package profiles
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/mypurecloud/platform-client-sdk-cli/build/gc/config"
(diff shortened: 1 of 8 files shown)


resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
```diff
@@ -67,23 +73,35 @@ func requestUserInput() config.Configuration {
 		environment = "mypurecloud.com"
 	}
 
-	for true {
+	fmt.Printf("Please provide either an Access Token, Client Credentials (Client ID and Client Secret) or both\n")
+	fmt.Printf("Note: If you provide an Access Token, this will be used over Client Credentials")
```
ronanwatkins, line 77: There needs to be a new line after this. I'm getting `Note: If you provide an Access Token, this will be used over Client CredentialsAccess Token:`

resources/sdk/clisdkclient/extensions/config/config.go
```diff
@@ -128,6 +139,10 @@ func applyEnvironmentVariableOverrides() {
 	if clientSecret != "" && ClientSecret == "" {
 		ClientSecret = clientSecret
 	}
+	accessToken := os.Getenv("GENESYS_CLOUD_ACCESS_TOKEN")
```
ronanwatkins, line 142: There should be no underscore between genesys and cloud like with the other environment variables

ronanwatkins:
Looks pretty good. A few changes are needed in the README to reflect this change. 

> All access to Genesys Cloud is provided via OAuth client credentials. Security is enforced via the OAuth client credentials.
Should be changed to indicate an access token is possible

The block containing the example config file should be updated to have another profile with just an access token specified.

The new environment variable needs to be added to `Environment variables`

The param override needs to be added to `Parameter overrides`

ronanwatkins:
@carnellj @timsmithgenesys Any opinions on calling this access token vs auth token? I see that in the docs for Client Credentials and Code Authorization they're used interchangeably.

timsmithgenesys:
@ronanwatkins It should be "access token" in all instances. This is the language used by the OAuth RFC: https://datatracker.ietf.org/doc/html/rfc6749#section-1.4

ronanwatkins (approved):

PR #232, opened 2021-09-21
Slurping in multiple files from a directory for "bulk" creating or updating data.

michaeljroddy:
Allowing users to provide a directory path to the --directory flag or to the -d short flag containing request bodies to "bulk" create or update data. For example, bulk creating or updating user data. If there is one file in the directory, the response will be a single object (similar to how the --file flag would work when provided with a single file for making a request). If there is more than one file (request bodies) in the directory then the response will be and array of objects returned from each individual request. If there are errors with any of the requests, the error objects will be concatenated onto the response where the successful response object would have been otherwise.

### DIFF
--- a/resources/sdk/clisdkclient/extensions/cmd/groups_members/addmembers.go
+++ b/resources/sdk/clisdkclient/extensions/cmd/groups_members/addmembers.go
@@ -1,6 +1,8 @@
 package groups_members
 
 import (
+	"bytes"
+	"encoding/gob"
 	"encoding/json"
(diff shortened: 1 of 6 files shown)


ronanwatkins (approved):
Works perfect. It would also be good to have an example of this in the Examples section. Nothing too fancy, just something to show that this flag can be used to create multiple users

carnellj (approved):

ronanwatkins (approved):
Very nice

PR #237, opened 2021-09-28
NO-JIRA - Made a small change to how YAML data is printed.

charliecon:
I added `---` to the line preceding a YAML response object. I think this could be more convenient for when a user wants to process multiple YAML response bodies stored in the same file as they would be separated.

### DIFF
--- a/resources/sdk/clisdkclient/extensions/utils/render.go
+++ b/resources/sdk/clisdkclient/extensions/utils/render.go
@@ -18,6 +18,7 @@ func Render(data string) {
 		if err != nil {
 			logger.Fatalf("Error converting JSON to YAML: %v\n", err)
 		}
+		fmt.Println("---")
 		fmt.Printf("%s", result)
 		return
 	}


michaeljroddy (commented):
Hi Charlie, this change will only work when outputting a single JSON object as YAML. If a users requests some data and the response returned is an array of JSON objects, this will not separate each individual object, the `---` will just be printed at the top of the response.  The JSONToYAML function will process the entire JSON response string as one. Try running `gc users list -a --outputformat=yaml` and you will see what I mean.

charliecon:
@michaeljroddy I hadn't considered that scenario yet but that functionality is actually ideal. In that case the triple hyphen will indicate the start of the document and each single hyphen will indicate an item in the list. There wouldn't be any issues processing that file. The change I'm proposing here would apply more to when someone redirects notifications to a file in yaml format. Without the triple hyphen each chunk will be thrown on top of each other with nothing separating them.

ronanwatkins (approved):

PR #241, opened 2021-10-11
Devengage 837 - Implementing Implicit Grant login for the CLI 

charliecon:


resources/sdk/clisdkclient/extensions/models/implicitWebpage.go
```diff
@@ -0,0 +1,48 @@
...
+func (w ImplicitWebpage) GetImplicitWebpage(clientID string, environment string, redirectURI string) string {
+	return fmt.Sprintf(AuthWebpage, clientID, redirectURI, environment)
+}
+
+const AuthWebpage = `
+<html>
+<head>
+    <title>Implicit Login</title>
+
+    <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
```
carnellj, line 18: @timsmithgenesys Couple of thoughts:

1.  Do we want to reference a third party CDN?  If the CDN is unavailable the user cant authenticate with the implicit grant?
2. GoLang now supports embedded resources.  Rather then putting a web page right here why not embedd the HTML and a local copy of the above javascript file right in the binary and distribute it with the binary?

Thoughts,
    John

resources/sdk/clisdkclient/extensions/restclient/tls/generate_cert.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if err := pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes}); err != nil {
+		return err
+	}
+	if err := certOut.Close(); err != nil {
+		return err
+	}
+	log.Print("wrote cert.pem\n")
+
+	keyOut, err := os.OpenFile(os.TempDir() + "key.pem", os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
ronanwatkins, line 86: Ah cool, I never knew about os.TempDir()

resources/sdk/clisdkclient/extensions/restclient/tls/generate_cert.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return err
+	}
+	if err := pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes}); err != nil {
+		return err
+	}
+	if err := certOut.Close(); err != nil {
+		return err
+	}
+	log.Print("wrote cert.pem\n")
```
ronanwatkins, line 84: If you want to log it would be best to use our own "logger" package

resources/sdk/clisdkclient/extensions/restclient/tls/generate_cert.go
```diff
@@ -0,0 +1,103 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"log"
+	"math/big"
+	"os"
+	"time"
+)
+
+var (
```
ronanwatkins, line 25: Looks like these are all constant vars so you can change var to const

resources/sdk/clisdkclient/extensions/restclient/tls/generate_cert.go
```diff
@@ -0,0 +1,103 @@
...
+	host     = "localhost"
+	validFor = 365*24*time.Hour
+	rsaBits  = 2048
+)
+
+func GenerateCerts() error {
+	var priv interface{}
+	var err error
+
+	priv, err = rsa.GenerateKey(rand.Reader, rsaBits)
```
ronanwatkins, line 35: Should this error be checked and returned if necessary?

charliecon: Yes! It must have been lost in the refactoring

resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
```diff
@@ -77,35 +104,94 @@ func requestUserInput() config.Configuration {
...
+		}
+	}
+
+	clientID, clientSecret = requestClientCreds(accessToken, grantType)
+
+	if grantType == ImplicitGrant {
+		redirectURI = requestRedirectURI()
+		fmt.Printf("Redirect URI: %s\n", redirectURI)
+		for true {
+			fmt.Print("Would you like to open a secure HTTP connection? [Y/N]: ")
```
ronanwatkins, line 131: This should be asked before they are asked for the redirect URI. And `open` should be switched to `use`

resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
```diff
@@ -77,35 +104,94 @@ func requestUserInput() config.Configuration {
...
+		fmt.Print("Client Secret (Optional): ")
+		fmt.Scanln(&secret)
+	}
+
+	return id, secret
+}
+
+func requestRedirectURI() string {
+	var inputPort string
+	redirectURI := "http://localhost:"
```
ronanwatkins, line 183: This should be HTTPS if they choose a secure connection.

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -243,6 +283,93 @@ func authorize(c config.Configuration) (models.OAuthTokenData, error) {
 	return *oAuthTokenResponse, nil
 }
 
+// Starts the local server at the redirect URI
+func startLocalServerFunc(c config.Configuration, authChannel chan models.OAuthToken) {
+	address := strings.Replace(c.RedirectURI(), "http://localhost", "", -1)
```
ronanwatkins, line 288: This code may need to be updated if the redirect URI is stored as `https://localhost`

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -243,6 +283,93 @@ func authorize(c config.Configuration) (models.OAuthTokenData, error) {
...
+			return
+		}
+
+		if strings.Contains(r.URL.String(), "/error/") || strings.Contains(r.URL.String(), "/access_token/") {
+			oAuthToken = getImplicitResponseDataFromURL(r.URL.String())
+			activePage = implicitWebpage.GetResponsePage(oAuthToken.Error)
+		}
+	})
+
+	if c.SecureLoginEnabled() {
```
ronanwatkins, line 323: There's some warning logging coming from the HTTP package when we use HTTPS. Call `log.SetOutput(ioutil.Discard)` before `http.ListenAndServe` to ignore it

ronanwatkins: I think you unresolved this by accident

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -176,6 +186,29 @@ func ReAuthenticate(c config.Configuration) (models.OAuthTokenData, error) {
...
 func authorize(c config.Configuration) (models.OAuthTokenData, error) {
+	authChannel := make(chan models.OAuthToken)
+
+	// Using implicit grant
+	if c.RedirectURI() != "" {
+		if c.SecureLoginEnabled() {
+			// Generate a self-signed X.509 certificate for a TLS server
+			err := tls.GenerateCerts()
+			if err != nil {
+				logger.Fatalf("Error generating self-signed certificate: %v\n", err)
```
ronanwatkins, line 197: This error should probably be returned

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -243,6 +282,93 @@ func authorize(c config.Configuration) (models.OAuthTokenData, error) {
...
+	var oAuthToken models.OAuthToken
+	implicitWebpage := models.ImplicitWebpage{}
+	// Set initial web page as one containing the javascript to perform implicit login
+	initialPage := implicitWebpage.GetImplicitWebpage(c.ClientID(), c.Environment(), c.RedirectURI())
+	activePage := initialPage
+
+	http.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		_, err := fmt.Fprintf(w, activePage)
+		if err != nil {
+			logger.Fatal(err)
```
ronanwatkins, line 309: Maybe add something to the log like "Error handling request: ${err}"

resources/sdk/clisdkclient/templates/README.md
```diff
@@ -24,7 +24,10 @@ If you are using the CLI in a pipeline or don't wish to set up a config file, se
 
 The "environment" value can be provided as the API base path such as `mypurecloud.com` or AWS region.
 
-To setup and configure your gc CLI run `gc profiles new` command and answer the questions.  If everything works correctly you should have a file created in your home directory called .gc/config.toml.  If for some reason you have problems with the `gc profiles new` command, you can manually create the config file by following the steps below.
+To setup and configure your gc CLI run `gc profiles new` command and answer the questions. If you choose not to use client credentials, you will be asked to enter a port number for the redirect URI that will be used to validate the client implicitly through a web browser.
```
ronanwatkins, line 27: You'll need to add a dedicated section for implicit grant. Indicate that they need to create an OAuth client with http://localhost:[PORT] or https://localhost:[PORT] and why they'd want to use implicit (user context)

resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
```diff
@@ -77,35 +104,100 @@ func requestUserInput() config.Configuration {
...
+			if strings.ToUpper(authChoice) == "Y" {
+				secureLoginEnabled = true
+				break
+			} else if strings.ToUpper(authChoice) == "N" {
+				secureLoginEnabled = false
+				break
+			}
+		}
+		redirectURI = requestRedirectURI(secureLoginEnabled)
+		fmt.Printf("Redirect URI: %s\n", redirectURI)
```
ronanwatkins, line 140: Print this out after they select whether they want a secure connection or not. Then it will be correct (http or https)

charliecon: The function `requestRedirectURI` will return the string with "http" or "https" depending on the value in the input param. Do you mean I should ask for the port, then if they want a secure connection, then print it out?

ronanwatkins: Yep, so it would look like this:
```
Profile Name [DEFAULT]: 
Environment [mypurecloud.com]: a
Note: If you provide an access token, this will take precedence over any authorization grant type.
Access Token (Optional): 
Select your authorization grant type.
        0. None
        1. Client Credentials
        2. Implicit Grant
Grant Type: 2
Client ID: a
Client Secret (Optional): 
Redirect URI port [8080]: 
Would you like to open a secure HTTP connection? [Y/N]: Y
Redirect URI: https://localhost:8080
Profile name DEFAULT already exists in the config file. Overwrite (Y/N): 
```

resources/sdk/clisdkclient/templates/restclient.mustache
```diff
@@ -243,6 +282,93 @@ func authorize(c config.Configuration) (models.OAuthTokenData, error) {
...
+// Starts the local server at the redirect URI
+func startLocalServerFunc(c config.Configuration, authChannel chan models.OAuthToken) {
+	var host string
+	if c.SecureLoginEnabled() {
+		host = "https://localhost"
+	} else {
+		host = "http://localhost"
+	}
+	address := strings.Replace(c.RedirectURI(), host, "", -1)
+	port := address[0:5]
```
timsmithgenesys, line 294: If you store RedirectURI as a URI (or at least parse it to one here), you won't have to come up with your own parsing logic to extract the different parts.

resources/sdk/clisdkclient/extensions/cmd/profiles/createprofile.go
```diff
@@ -77,35 +104,100 @@ func requestUserInput() config.Configuration {
...
+	id := ""
+	secret := ""
+
+	if grantType == ClientCredentials {
+		if accessToken != "" {
+			fmt.Print("Client ID: ")
+			fmt.Scanln(&id)
+
+			fmt.Print("Client Secret: ")
+			fmt.Scanln(&secret)
```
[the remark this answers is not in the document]
ronanwatkins: Do you mean to avoid it showing up on the terminal as it's entered? Something like this question? https://stackoverflow.com/questions/2137357/getpasswd-functionality-in-go
If so looks like there's some good solutions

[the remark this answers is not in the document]
timsmithgenesys: I read the code without first consuming a sufficient amount of coffee; you're not logging the value like I said. Hiding the input value would be excellent if it's not too difficult to implement. Using a user input library like that probably has some other fringe benefits as well.

ronanwatkins (approved):

PR #257, opened 2021-12-06
Adding ability to pass a github url to the transform flag + unit tests

charliecon:


resources/sdk/clisdkclient/extensions/transform_data/transform_data.go
```diff
@@ -70,4 +85,47 @@ func handleParse(t *template.Template, err error) *template.Template {
...
-}
\ No newline at end of file
+}
+
+func pullRemoteFileToTmpDir(url string) error {
+	rawURL := getFullUrlToRawFile(url)
+	var cmd *exec.Cmd
+
+	// curl request to pull file to current directory
+	cmd = exec.Command("curl", "-LJO", rawURL)
```
ronanwatkins, line 95: Best to use http.Get for downloading. You can also avoid writing a temp file using this method because the file contents would be in the resp.Body variable

resources/sdk/clisdkclient/extensions/transform_data/transform_data.go
```diff
@@ -42,8 +48,26 @@ func convertArrayOfObjectsToMap(data string) []map[string]interface{} {
 
 func ProcessTemplateFile(mp interface{}) string {
 	path := []string{TemplateFile}
-	name := filepath.Base(path[0])
-	tmpl := handleParse(template.New(name).Funcs(sprig.TxtFuncMap()).ParseFiles(path...))
+	fileName = filepath.Base(path[0])
+
+	if strings.Contains(TemplateFile, "github") {
```
ronanwatkins, line 53: See if you can make a regex that matches the accepted URL patterns. At the moment you could cod it with something like github.gotmpl

resources/sdk/clisdkclient/extensions/transform_data/transform_data.go
```diff
@@ -42,8 +48,26 @@ func convertArrayOfObjectsToMap(data string) []map[string]interface{} {
...
+		rawURL := getFullUrlToRawFile(TemplateFile)
+
+		resp, err := http.Get(rawURL)
+		if err != nil {
+			logger.Fatal(err)
```
ronanwatkins, line 58: It might be more helpful to preprend this and the following error with a message. Something like "Error downloading template file:" for this one would be helpful

PR #259, opened 2021-12-21
Target .NET framework version 4.5.2 as well as netstandard 2.0

ronanwatkins:


resources/sdk/pureclouddotnet/scripts/publishNuget.sh
```diff
@@ -40,14 +40,17 @@ echo '<?xml version="1.0" encoding="utf-8"?>'\
...
+'				<dependency id="RestSharp" version="106.3.1" />'\
+'				<dependency id="ini-parser" version="3.4.0" />'\
+'			</group>'\
+'			<group targetFramework=".NETFramework4.8" />'\
 '		</dependencies>'\
 '	</metadata>'\
 '	<files>'\
-"		<file src=\"$NAMESPACE.dll\" target=\"lib/netstandard2.0/$NAMESPACE.dll\" />"\
-"		<file src=\"$NAMESPACE.xml\" target=\"lib/netstandard2.0/$NAMESPACE.xml\" />"\
+"		<file src=\"$NAMESPACE.dll\" target=\"lib/net48/$NAMESPACE.dll\" />"\
```
timsmithgenesys, line 52: Could we _add_ support for a newer version without dropping the existing one? Just thinking about backwards compatibility and also the ability to add multiple support targets (if we want to). https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks

timsmithgenesys (approved):

resources/sdk/pureclouddotnet/scripts/publishNuget.sh
ronanwatkins: I've managed to add support for 4.5.2 while keeping the original support. I can't go lower than 4.0 because we're using the `await` keyword and support for that was added since 4.0.
I'd add support for newer frameworks but mono would have to be updated on jenkins and I'm not going to open that can of worms unless absolutely necessary.

PR #310, opened 2022-03-01
DEVENGAGE-826 - API Data generation for SDK Invocations

diemenschmaschine92:
Adding combineApiDataFiles.js script and configuration to SDKs


resources/sdk/clisdkclient/scripts/combineApiDataFiles.js
```diff
@@ -0,0 +1,26 @@
+const fs = require('fs');
```
timsmithgenesys, line 1: Can this script be moved to a shared location instead of duplicated in each folder? This kind of duplication is very likely to result in the scripts getting out of sync at some point in the future.

resources/sdk/clisdkclient/templates/api_json.mustache
```diff
@@ -0,0 +1,14 @@
+{
```
timsmithgenesys, line 1: Is it possible to move this template to a common location as well?

ronanwatkins: Could probably achieve this by storing it in `resources/templates` and passing it in as another `-t` flag in [builder.js](https://github.com/purecloudlabs/platform-client-sdk-common/blob/master/modules/builder.js#L370)

diemenschmaschine92: I was able to move the script to a common folder, but wasn't able to pass multiple -t options.  I found that the first template folder passed in was used, but not both.

ronanwatkins: Would it be possible to store it in `resources/templates` and keep a sym link to it in each SDK template directory? If that doesn't work then we'll have to leave it as is.

resources/sdk/webmessagingjava/templates/api_json.mustache
```diff
@@ -0,0 +1 @@
+/Users/jshaw/genesys_src/repos/purecloudlabs/platform-client-sdk-common/resources/templates/api_json.mustache
```
ronanwatkins, line 1: Could you do it as a relative sym link?

diemenschmaschine92: Ah I see what you mean.  This is my first exposure to symlinks so I'm fumbling a bit. Will do.

ronanwatkins (approved):

PR #315, opened 2022-03-02
[DEVENGAGE-990] Python SDK - updating dependencies & enabling dependabot

charliecon:


resources/sdk/purecloudpython/templates/requirements.txt
```diff
@@ -0,0 +1,6 @@
+certifi >= 14.05.14
```
ronanwatkins, line 1: You'll have to update PythonClientCodegen.java in swagger codegen and change requirements.mustache to requirements.txt

charliecon: PR is up - https://github.com/purecloudlabs/swagger-codegen/pull/49

PR #330, opened 2022-03-17
SHYRKA-1344 add method to handle error on the websocket connection

gartcimore:
add method to handle error on the websocket connection
add method to check status of input and output channel of the websocket
add option in connect to set the connection timeout

changed formating of the file


gartcimore:
@Jm-47 if you have any comments

PR #371, opened 2022-05-13
[DEVENGAGE-1001] Supporting HEAD requests on the CLI

charliecon:


ronanwatkins (approved):

resources/sdk/clisdkclient/extensions/services/commandservice.go
```diff
@@ -445,6 +446,10 @@ func (c *commandService) Delete(uri string) (string, error) {
 	return c.upsert(http.MethodDelete, uri, "")
 }
 
+func (c *commandService) Head(uri string) (string, error) {
+	return c.upsert(http.MethodHead, uri, "")
```
ronanwatkins, line 450: You could rename upsert to something like invoke or callAPI now since it can be used for GET like operations. Also I'm pretty sure you could change the `func (c *commandService) Get(uri string) (string, error)` function to call upsert like this to tidy things up.

PR #386, opened 2022-06-03
fixing java sdk build failure

michaeljroddy:
dependabot updated testng from version 7.5 to version 7.6.0 recently. Looks like the new testng version requires jdk 11 but we are using jdk 8. so rolling back testng to most recent version where it works.


ronanwatkins:
Nice one. Really need to get a pipeline set up to avoid this stuff.
Could you configure dependabot to ignore this dependency so it doesn't keep opening PRs for it: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#disabling-dependabot-version-updates

PR #398, opened 2022-06-28
Adding a script to replace properties in the pom

ronanwatkins:
Will resolve this issue: https://developer.genesys.cloud/forum/t/could-not-resolve-latest-platform-client-v2/15310

Gradle doesn't like pom files with properties for the artefact version so this script will replace them will actual values


timsmithgenesys:
@ronanwatkins do you know why the pom isn't a mustache file? Seems like if it were made a mustache file it could have variables in it and wouldn't need the script. https://github.com/purecloudlabs/platform-client-sdk-common/blob/master/resources/sdk/purecloudjava/templates/pom.xml

charliecon (approved):

timsmithgenesys (approved):

PR #436, opened 2022-09-27
[DEVENGAGE-1429] --filtercondition flag for the CLI list command 

charliecon:


resources/sdk/clisdkclient/extensions/utils/filter.go
```diff
@@ -0,0 +1,270 @@
...
+		allMatchedObjects []interface{}
+		currentObject     interface{}
+		match             bool
+		err               error
+	)
+
+	for _, object := range returnedObjects {
+		currentObject = object
+		for i, k := range keys {
+			if currentObjectMap, ok := currentObject.(map[string]interface{}); ok {
```
carnellj-genesys, line 83: Hi Charlie,

I would refactor the code below to get rid of the if..then..else.  Properly structured go will do a | line down from each if..then statement.  So you check something and grab the data and return rather then doing a nest of if..then statements.

resources/sdk/clisdkclient/extensions/utils/filter.go
```diff
@@ -0,0 +1,270 @@
...
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func fieldMatchesValue(jsonValue interface{}, cliInputValue string, operator string) (bool, error) {
+	if fieldValueStr, ok := jsonValue.(string); ok {
+		return compareStrings(fieldValueStr, cliInputValue, operator)
+	} else if fieldValueBool, ok := jsonValue.(bool); ok {
```
carnellj-genesys, line 135: See my earlier comments about if..then

carnellj-genesys (approved):
Just a couple of small items, but otherwise looks good

resources/sdk/clisdkclient/extensions/utils/filter.go
charliecon: @carnellj-genesys  done ✅

PR #452, opened 2022-10-19
bug/DEVENGAGE-1420 varType bug in CLI

dginty4:


resources/sdk/clisdkclient/templates/api.mustache
```diff
@@ -73,6 +73,11 @@ var {{operationId}}Cmd = &cobra.Command{
 		if len(queryParams) > 0 {
 			urlString = fmt.Sprintf("%v?", path)
 			for k, v := range queryParams {
+                if k == "varType" {
+                    k = "type"
+                    urlString += fmt.Sprintf("%v=%v&", url.QueryEscape(strings.TrimSpace(k)), url.QueryEscape(strings.TrimSpace(v)))
+                    continue
+                }
```
charliecon, line 80: Looks like the indentation is a bit off. You can fix this by opening this file in vs code and tabbing this block in and out again

carnellj-genesys (approved):

resources/sdk/clisdkclient/templates/api.mustache
```diff
@@ -73,6 +73,11 @@ var {{operationId}}Cmd = &cobra.Command{
 		if len(queryParams) > 0 {
 			urlString = fmt.Sprintf("%v?", path)
 			for k, v := range queryParams {
+				if k == "varType" {
+					k = "type"
+					urlString += fmt.Sprintf("%v=%v&", url.QueryEscape(strings.TrimSpace(k)), url.QueryEscape(strings.TrimSpace(v)))
+					continue
```
charliecon, line 79: I don't think there's any need for lines 78 and 79, unless I'm missing something

michaeljroddy (commented):
another way to do this might be to just check if the `urlString` contains `varType` (after the for loop) and if it does, replace it with `type` like this:

```
if strings.Contains(urlString, "varType") {
        urlString = strings.Replace(urlString, "varType", "type", -1)
}
```

@charliecon @dginty4 thoughts?

charliecon:
> another way to do this might be to just check if the `urlString` contains `varType` (after the for loop) and if it does, replace it with `type` like this:
> 
> ```
> if strings.Contains(urlString, "varType") {
>         urlString = strings.Replace(urlString, "varType", "type", -1)
> }
> ```
> 
> @charliecon @dginty4 thoughts?

Yeah we're probably splitting hairs but I suppose less nesting is always better

dginty4:
> > another way to do this might be to just check if the `urlString` contains `varType` (after the for loop) and if it does, replace it with `type` like this:
> > ```
> > if strings.Contains(urlString, "varType") {
> >         urlString = strings.Replace(urlString, "varType", "type", -1)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @charliecon @dginty4 thoughts?
> 
> Yeah we're probably splitting hairs but I suppose less nesting is always better

I might just change it to this, I think its simpler

charliecon (approved):

michaeljroddy (approved):

PR #474, opened 2022-11-17
Devengage 1323 purecloudgo remove omitempty optional properties

michaeljroddy:
To use this solution, you can set the `PatchConfiguration` directly on the `Configuration` object or you can use the `platformclientv2.SetPatchConfiguration()` function.

#### Example 1: Setting `PatchConfiguration` directly on the `Configuration` object:

Set `RemoveOmitEmpty` to true and provide a slice of strings containing the APIs you want this logic to affect on.

```
config := platformclientv2.GetDefaultConfiguration()
patchConfig := &platformclientv2.PatchConfiguration{RemoveOmitEmpty: true, APIs: []string{"/api/v2/journey/", "/api/v2/users/"}}
config.PatchConfiguration = patchConfig
```

In the example above, the `omitempty` tags will be removed from optional properties in post bodies on `PATCH /api/v2/journey` endpoints and `PATCH /api/v2/users` endpoints when making a request.

#### Example 2: Setting `PatchConfiguration` using `platformclientv2.SetPatchConfiguration()`:

Or you can set the `PatchConfiguration` anywhere in your code using the `SetPatchConfiguration()` function.

For example, assume a default configuration has been set (so `PatchConfiguration == nil`) and you want to set the `PatchConfiguration` at a later stage. You can do the following:

```
patchConfig := &platformclientv2.PatchConfiguration{RemoveOmitEmpty: true, APIs: []string{"/api/v2/journey/"}}
platformclientv2.SetPatchConfiguration(patchConfig)
```
Similarly, set `RemoveOmitEmpty` to true and provide a slice of strings containing the APIs you want this logic to affect on.

The function `platformclientv2.SetPatchConfiguration()` will set or override the current `PatchConfiguration`.

**Note:** There are some print statements in the code as i find them useful for testing/debugging etc..

### DIFF
--- a/resources/sdk/purecloudgo/templates/apiclient.mustache
+++ b/resources/sdk/purecloudgo/templates/apiclient.mustache
@@ -17,8 +17,8 @@ import (
 
 // APIClient provides functions for making API requests
 type APIClient struct {
-	client             retryablehttp.Client
-	configuration      *Configuration
+	client        retryablehttp.Client
+	configuration *Configuration
 }
 
 // NewAPIClient creates a new API client
@@ -118,7 +118,6 @@ func (c *APIClient) CallAPI(path string, method string,
 	// Set post body
 	if postBody != nil {
 		request.Header.Set("Content-Type", "application/json")
-
 		if c.configuration.PatchConfiguration != nil && method == "PATCH" && c.configuration.PatchConfiguration.RemoveOmitEmpty && pathContainsAPI(path, c.configuration.PatchConfiguration.APIs) {
 			processedModel, _ := processModel(postBody)
 			fmt.Println("processed model")
@@ -292,9 +291,9 @@ func copy(data []byte, v interface{}) {
 func processModel(model interface{}) ([]byte, error) {
 	value := reflect.ValueOf(model)
 	// If it's an interface or a pointer, unwrap it.
-    if value.Kind() == reflect.Ptr && value.Elem().Kind() == reflect.Struct {
-        value = value.Elem()
-    }
+	if value.Kind() == reflect.Ptr && value.Elem().Kind() == reflect.Struct {
+		value = value.Elem()
+	}
 	t := value.Type()
 	sf := make([]reflect.StructField, 0)
 	for i := 0; i < t.NumField(); i++ {
--- a/resources/sdk/purecloudgo/templates/apiclient.mustache
+++ b/resources/sdk/purecloudgo/templates/apiclient.mustache
@@ -17,8 +17,8 @@ import (
 
 // APIClient provides functions for making API requests
 type APIClient struct {
-	client        retryablehttp.Client
-	configuration *Configuration
+	client             retryablehttp.Client
+	configuration      *Configuration
 }
 
 // NewAPIClient creates a new API client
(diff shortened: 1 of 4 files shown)


istvan-bihari-genesys:
I like it. I plan to validate at https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/309 PR.

resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -118,8 +118,15 @@ func (c *APIClient) CallAPI(path string, method string,
 	// Set post body
 	if postBody != nil {
 		request.Header.Set("Content-Type", "application/json")
-		j, _ := json.Marshal(postBody)
-		request.SetBody(ioutil.NopCloser(bytes.NewReader(j)))
+		if c.configuration.PatchConfiguration != nil && method == "PATCH" && c.configuration.PatchConfiguration.RemoveOmitEmpty && pathContainsAPI(path, c.configuration.PatchConfiguration.APIs) {
+			processedModel, _ := processModel(postBody)
```
carnellj-genesys, line 122: A couple of small things here:

1.  You have a couple of fmt.Println.  These should be logging debug statements.
2. Also you might want to explicitly call out that patch overrides are present before you process the model.  Otherwise, if a customer or support is debugging why their fields are overridden they can explicitly see if it is because the patch override models fields have been defined
3. Also you are ignoring the error here.  You should handle this gracefully and log the error and breakout out of the code.  With the way you have written this now the processedModel value will be NULL and you will be setting the SetBody with a nil value, which may cause a panic later on down the line.  

In go errors should be explicitly handled and ignoring an error value should be the exception instead of the norm.

resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -278,4 +285,36 @@ func copy(data []byte, v interface{}) {
 	dataS := string(data)
 
 	reflect.ValueOf(v).Elem().Set(reflect.ValueOf(&dataS))
+}
+
+// process the model to remove omitempty tags on optional properties
+func processModel(model interface{}) ([]byte, error) {
+	value := reflect.ValueOf(model)
```
carnellj-genesys, line 292: Super nice.  Good job figuring out the tags API.  This might be useful later on in other parts of CX as Code

resources/sdk/purecloudgo/templates/configuration.mustache
```diff
@@ -60,6 +61,17 @@ const (
...
 )
 
+func SetPatchConfiguration(patchConfiguration *PatchConfiguration) {
+	instance.PatchConfiguration = &PatchConfiguration{RemoveOmitEmpty: patchConfiguration.RemoveOmitEmpty, APIs: patchConfiguration.APIs}
+	fmt.Println("PatchConfiguration set")
+	fmt.Printf("%#v\n", instance.PatchConfiguration)
+} 
+
+type PatchConfiguration struct {
+	RemoveOmitEmpty bool     `json:"removeOmitEmpty,omitempty"`
```
carnellj-genesys, line 71: Wondering if this is redundant.  You can already determine if you need to remove a field by whether or not the array has any values.  If the display is empty then you do have not anything to omit.  Removing this field would. make the if condition when determining if you want to remove it much easier to carry out.

carnellj-genesys (changes requested):
Overall really nice job Michael.  I think this is a clean solution that maintains backward compatibility and provides a mechanism to bypass this problem.  Also, keep in mind 2 things:

1.  Lets make sure we update the documents.
2. Make an announcement in the SDK chat room to let internal SDKs users know a change has gone out.  The change shouldn't cause backward compatibility issues but always good to let them know we did something like this as the testers are our best canaries.  I would let it back for a couple of days internally and then deploy early next week.

timsmithgenesys:
superseded by https://github.com/purecloudlabs/platform-client-sdk-common/pull/505

PR #489, opened 2022-12-12
Removing symlink for webmessagingsdk api_json template

mark-siggins-genesys:
The other symlinks have already been reverted and this is causing internal build failures.

The other reverts are here - https://github.com/purecloudlabs/platform-client-sdk-common/commit/42e2183b90179b50bf49b8c04210f1875da6789d


mark-siggins-genesys:
@diemenschmaschine92 - this is the change we spoke about

charliecon (approved):

PR #499, opened 2023-01-03
Refactor UpgradeSdkEnumConverter to cache the result instead of using reflection each time it reads or writes JSON

RobinsonWM:
I was profiling an application that uses the Platform Client SDK (but also does many other things,) and I noticed that 25% of objects allocated on the .NET managed heap were coming from `UpgradeSdkEnumConverter`.  It uses reflection to map strings to enum values.

The implementation proposed on this PR caches the mapping of strings to enum values on the first run for each enum type, and then subsequent runs don't allocate any objects on the managed heap.  I didn't intend to make any changes to the functionality, so I added some new unit tests that pass with both the old and proposed implementations.

I wrote some benchmarks with BenchmarkDotNet to quantify the performance impact of the proposed change:
* `ReadJson` - just calls `UpgradeSdkEnumConverter.ReadJson()` (the method I optimized)
* `DeserializeEnum` - calls `JsonConvert.DeserializeObject<Conversation.StateEnum>("\"transmitting\"")`
* `DeserializeConversation` - calls `JsonConvert.DeserializeObject<Conversation>()` with a normal conversation object

The results show no allocation when `ReadJson()` is used alone, and a ~50% reduction in garbage collections when it is used in the context of deserializing a larger object.

#### Before fix:

|                  Method |         Mean |       Error |      StdDev |    Gen0 |   Gen1 | Allocated |
|------------------------ |-------------:|------------:|------------:|--------:|-------:|----------:|
|                ReadJson |   3,181.5 ns |    18.39 ns |    27.53 ns |  0.2174 |      - |    1040 B |
|         DeserializeEnum |  13,573.1 ns |   118.21 ns |   173.27 ns |  1.4038 | 0.0153 |    6617 B |
| DeserializeConversation | 306,949.0 ns | 2,066.34 ns | 3,092.80 ns | 20.9961 | 1.9531 |  100488 B |

#### After fix:

|                  Method |         Mean |        Error |       StdDev |   Gen0 |   Gen1 | Allocated |
|------------------------ |-------------:|-------------:|-------------:|-------:|-------:|----------:|
|                ReadJson |     36.65 ns |     0.602 ns |     0.863 ns |      - |      - |         - |
|         DeserializeEnum |    732.11 ns |    17.020 ns |    25.474 ns | 0.6113 | 0.0057 |    2880 B |
| DeserializeConversation | 93,841.69 ns | 1,260.671 ns | 1,886.914 ns | 9.5215 | 0.8545 |   44960 B |

>  Mean      : Arithmetic mean of all measurements
>  Error     : Half of 99.9% confidence interval
>  StdDev    : Standard deviation of all measurements
>  Gen0      : GC Generation 0 collects per 1000 operations
>  Gen1      : GC Generation 1 collects per 1000 operations
>  Allocated : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
>  1 ns      : 1 Nanosecond (0.000000001 sec)

### DIFF
--- a/resources/sdk/pureclouddotnet/config.json
+++ b/resources/sdk/pureclouddotnet/config.json
@@ -78,6 +78,18 @@
       ],
       "compileScripts": [
         {
+          "type": "node",
+          "path": "processExtensions.js",
+          "args": [
+            "${COMMON_ROOT}/resources/sdk/pureclouddotnet/extensions-test",
+            "${SDK_REPO}/build/src/${DOTNET_NAMESPACE}.Tests",
+            {
+              "$ref": "#/settings/namespace"
+            }
+          ],
+          "failOnError": true
+        },
+        {
           "type": "shell",
           "path": "compile.sh",
           "args": [
--- a/resources/sdk/pureclouddotnet-guest/extensions/Client/UpgradeSdkEnumConverter.cs
+++ b/resources/sdk/pureclouddotnet-guest/extensions/Client/UpgradeSdkEnumConverter.cs
@@ -1,99 +1,105 @@
 using System;
-using System.Linq;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Reflection;
 using System.Runtime.Serialization;
 using Newtonsoft.Json;
 
 namespace {{=it.packageName}}.Client
 {
-    class UpgradeSdkEnumConverter : JsonConverter
+    internal class UpgradeSdkEnumConverter : JsonConverter
     {
         // Inspired by http://stackoverflow.com/questions/22752075/how-can-i-ignore-unknown-enum-values-during-json-deserialization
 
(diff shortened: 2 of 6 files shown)


resources/sdk/pureclouddotnet/config.json
```diff
@@ -77,6 +77,18 @@
         }
       ],
       "compileScripts": [
```
RobinsonWM, line 79: This is a bit of a weird place to add unit tests, but it seemed like the least intrusive change.
The other tests were handled here, but I think it would require builds to be done in a specific order to avoid breaking if I added the new test file here: https://github.com/purecloudlabs/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages/PureCloudDotNetClientCodegen.java#L81

carnellj-genesys (approved):

charliecon (approved):

PR #505, opened 2023-01-23
DEVENGAGE-1323 field serialization opt in

timsmithgenesys:
Adds a helper function to models in the Go SDK to manually control which properties are serialized, regardless of their value (including nil). Example:

```go
	patchBody := platformclientv2.Patchactionmap{}
	patchBody.SetField("Id", platformclientv2.String("777-777"))
	patchBody.SetField("Version", platformclientv2.Int(7))
	patchBody.SetField("DisplayName", nil)

	patchAction := platformclientv2.Patchaction{}
	patchAction.SetField("ActionTargetId", platformclientv2.String("12345"))
	patchBody.SetField("Action", &patchAction)
```

yields:

```json
{
  "action": {
    "actionTargetId": "12345"
  },
  "displayName": null,
  "id": "777-777",
  "version": 7
}
```

This allows applications to explicitly construct request objects that will be serialized exactly the way they are created, ignoring the default serialization rules (i.e. not distinguishing between a field not being set at all vs. being set with a nil value).

### DIFF
--- a/resources/sdk/purecloudgo/extensions/platformclientv2_test.go
+++ b/resources/sdk/purecloudgo/extensions/platformclientv2_test.go
@@ -142,6 +142,128 @@ func TestValueSerialization(t *testing.T) {
 	}
 }
 
+func TestCustomSerialization1(t *testing.T) {
+	patchBody := Patchactionmap{}
+
+	// Use SetField to trigger custom serialization
+	patchBody.SetField("Id", String("111-111"))
+
+	// These manually set fields won't appear in the output when SetField is used
+	patchBody.Version = Int(1)
+	patchBody.DisplayName = String("manually set display name")
+
+	// Validate serialization
+	j, err := json.Marshal(patchBody)
+	if err != nil {
+		t.Error(err)
+	}
+	s := string(j)
+	expected := `{"id":"111-111"}`
+	if s != expected {
+		t.Log("Patchactionmap did not serialize correctly")
+		t.Logf("Expected: %v", expected)
+		t.Logf("Actual:   %v", s)
+		t.FailNow()
+	}
+}
+
+func TestCustomSerialization2(t *testing.T) {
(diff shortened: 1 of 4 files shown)


resources/sdk/purecloudgo/templates/model.mustache
```diff
@@ -3,20 +3,77 @@ package {{packageName}}
...
+	}
+
+	// Add field to set field names list
+	if o.SetFieldNames == nil {
+		o.SetFieldNames = make(map[string]bool)
+	}
+	o.SetFieldNames[field] = true
+}
+
+func (o {{classname}}) MarshalJSON() ([]byte, error) {
```
timsmithgenesys, line 43: I changed the struct attachment of the MarshalJSON field from `(o *{{classname}})` to just `(o {{classname}})`. The custom serializer wasn't getting called in my testing with the existing signature. I'm not sure why it was that way and if changing it will have any negative impacts. The SDK's tests still pass.

resources/sdk/purecloudgo/extensions/platformclientv2_test.go
```diff
@@ -142,6 +142,128 @@ func TestValueSerialization(t *testing.T) {
...
+	patchAction.MediaType = nil
+	patchBody.SetField("Action", &patchAction)
+
+	// Validate serialization
+	j, err := json.Marshal(patchBody)
+	if err != nil {
+		t.Error(err)
+	}
+	s := string(j)
+	expected := `{"action":{"actionTargetId":"12345","isPacingEnabled":false},"displayName":null,"id":"333-333","version":3}`
```
timsmithgenesys, line 224: This result demonstrates mixing custom and default serialization. The root object contains `"displayName":null` because it was set using SetField but the `action` object does not contain `mediaType` because default serialization omits `nil` values.

charliecon (approved):

PR #513, opened 2023-02-09
webmessaging - add startNew param for configure session after disconnect event

gartcimore:
add startNew param for configure session after disconnect event for webmessaging SDK. This param is to be used after a Presence event of type disconnect is received, when feature disconnect is enabled and configure to set sessions to read-only after such event
Also adding methods to check if message contains events


mark-siggins-genesys (approved):
Useful additions.

charliecon (approved):

PR #521, opened 2023-02-27
DEVENGAGE-1610 Axios proxy bug

dginty4:


resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -1,6 +1,7 @@
 import { default as axios } from 'axios';
 import Configuration from './configuration.js';
 import { default as qs } from 'qs';
+const { HttpsProxyAgent } = require('hpagent')
```
dginty4, line 4: I couldn't get the import to work like the other imports, anybody know how I could?

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -275,16 +276,21 @@ class ApiClient {
...
-
 			const headers = {
 				'Authorization': `Basic ${authHeader}`
 			}
+            let httpsAgent;
+            if (this.proxy !== undefined){
+                httpsAgent = new HttpsProxyAgent({
+                    proxy : `${this.proxy.protocol}://${this.proxy.host}:${this.proxy.port}`
+                });
+            }
```
dginty4, line 287: I also can't seem to get the formatting right, it looks right on any editor I open it on except here

timsmithgenesys (commented):
Did we get a permanent proxy set up to be able to add tests? If so, tests should be added. If not, tests should be added when the proxy is deployed and accessible during the build.

carnellj-genesys (changes requested):
1.  Didn't have any technical comments besides what Tim pointed out.
2. I don't see any tests here.  I would like to see a test case using the target proxy.  While manual tests are good, automated tests will help us find stuff later if somebody changes something again.

Nice work on this Declan.

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
dginty4: Thanks Tim. `import { HttpsProxyAgent } from 'hpagent'` doesn't seem to work so I'll have to leave it as is

dginty4:
Would someone be able to build the sdk local? I think the security group for the proxy servers EC2 instance might only allow my IP through but I'm not 100%

carnellj-genesys (approved):

PR #597, opened 2023-05-18
bug/DEVENGAGE-1857 - Removing deprecation warnings from python

dginty4:


dginty4:
I'm going to look at a different way to add deprecation warning to python,. I am going to remove them for now so the sdk builds again

resources/sdk/purecloudpython/templates/api.mustache
```diff
@@ -60,8 +59,6 @@ class {{classname}}(object):
                 config.api_client = ApiClient()
             self.api_client = config.api_client
 {{#operation}}
-
-    {{#isDeprecated}}@deprecated("{{operationId}} is deprecated"){{/isDeprecated}}
     def {{operationId}}(self, {{#sortParamsByRequiredFlag}}{{#allParams}}{{#required}}{{paramName}}: {{#vendorExtensions}}{{{x-dataType}}}{{/vendorExtensions}}, {{/required}}{{/allParams}}{{/sortParamsByRequiredFlag}}**kwargs) -> {{#returnType}}{{#vendorExtensions}}{{{x-returnType}}}{{/vendorExtensions}}{{/returnType}}{{^returnType}}None{{/returnType}}:
```
HemanthDogiparthi12, line 62: if you are having problem with @deprecated. 
may be you could implement something like this and try ?

import warnings
from functools import wraps

def deprecated(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        warnings.warn(f"{{operationId}} is deprecated", DeprecationWarning)
        return func(*args, **kwargs)
    return wrapper

@deprecated
def {{operationId}}():
------

the deprecated decorator wraps the  deprecated function. When it is called, it will generate a deprecation warning  message

HemanthDogiparthi12 (approved):

resources/sdk/purecloudpython/templates/api.mustache
dginty4: Thanks Hemanth I'll try that

PR #631, opened 2023-07-24
Devengage 2019

HemanthDogiparthi12:
Previously , Client has this capability to modify underlying RestClient Properties that SDK uses for making their Platform API calls. Since the upgrade, RestClient object has become lot more light weight with most of the responsibilities transferred to RestClientOptions. Made modifications to make sure any changes to the Rest Behaviour can be only done via exposed SDK setter/getter methods of Configuration and ClientOptions classes and propagate their properties while making the API calls via RestSharp.

### DIFF
--- a/resources/sdk/pureclouddotnet/templates/ApiClient.mustache
+++ b/resources/sdk/pureclouddotnet/templates/ApiClient.mustache
@@ -113,7 +113,7 @@ namespace {{packageName}}.Client
         /// Gets or sets the RestClient.
         /// </summary>
         /// <value>An instance of the RestClient</value>
-        public RestClient RestClient { get; set; }
+        private RestClient RestClient { get; set; }
 
         private RetryConfiguration retryConfig;
         public RetryConfiguration RetryConfig { get; set; }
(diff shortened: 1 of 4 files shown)


resources/sdk/pureclouddotnet/templates/ApiClient.mustache
```diff
@@ -113,7 +113,7 @@ namespace {{packageName}}.Client
         /// Gets or sets the RestClient.
         /// </summary>
         /// <value>An instance of the RestClient</value>
-        public RestClient RestClient { get; set; }
+        private RestClient RestClient { get; set; }
```
carnellj-genesys, line 116: I am a little concerned about restricting access to this.  Usually, we would let the rest client in an API be exposed so that if there is a corner case that the API can not handle, the customer might still be able to do the work with RestClient.  Is there a specific reason why we decided to change this from public to private?  

Thanks,
     John

carnellj-genesys (approved):

charliecon (approved):

PR #635, opened 2023-08-02
Shyrka 2103 include application details at connection

gartcimore:
Adding a way to have like an user-agent signature when connecting to the websocket for webmessaging.
Default connecting method will not fill this and the default will be used ("WebMessagingSDK-{{version of the sdk}}")


carnellj-genesys (approved):

HemanthDogiparthi12 (approved):

gartcimore:
This seems approved, do we need something more before merging it ?

PR #656, opened 2023-09-04
Fix GC error

HemanthDogiparthi12:


carnellj-genesys (approved):
What ticket is this associate with Hemanth?

PR #662, opened 2023-09-08
NO-JIRA: Backing out the version change I made

carnellj-genesys:


charliecon (approved):

resources/sdk/purecloudgo/templates/go.mod.mustache
```diff
@@ -1,4 +1,4 @@
-module github.com/mypurecloud/platform-client-sdk-go
+module github.com/mypurecloud/platform-client-sdk-go/{{major-version}}
```
charliecon, line 1: Is it {{major-version}} or {{majorVersion}}. ?

charliecon: After looking at git blame ,it used to be {{majorVersion}} camel case

PR #666, opened 2023-09-18
NO-JIRA IOS install error

dginty4:


resources/sdk/purecloudios/templates/Podspec.mustache
```diff
@@ -8,7 +8,7 @@ Pod::Spec.new do |s|
   s.source = { :git => 'https://github.com/MyPureCloud/platform-client-sdk-ios.git', :tag => '{{podVersion}}' }{{#podAuthors}}
   s.authors = '{{podAuthors}}'{{/podAuthors}}{{#podSocialMediaURL}}
   s.social_media_url = '{{podSocialMediaURL}}'{{/podSocialMediaURL}}{{#podDocsetURL}}
-  s.docset_url = '{{podDocsetURL}}'{{/podDocsetURL}}
+  #s.docset_url = '{{podDocsetURL}}'{{/podDocsetURL}}
```
dginty4, line 11: This line causes an issue when installing the IOS sdk, the build fails when it's removed so I commented it out

HemanthDogiparthi12 (approved):

PR #839, opened 2024-05-29
Bug/devtooling 599

kavinbalagen:
override transitive dependency version


carnellj-genesys (approved):

carnellj-genesys:
@HemanthDogiparthi12 @dginty4 - Please review before merging.  For some reason, I can add you as reviewers

HemanthDogiparthi12 (commented):
@kavinbalagen if possible , Check in webmessaging as well, if this dependency is used and need to be updated.

PR #914, opened 2024-09-23
Devtooling-664: Added Gateway Config to goSDK

BrianMoyles:


resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -106,20 +147,8 @@ func (c *APIClient) CallAPI(path string, method string,
 	fileBytes []byte,
 	pathName string) (*APIResponse, error) {
 
-	// Build request URL w/query params
 	urlString := path + "?"
-	if queryParams != nil {
```
HemanthDogiparthi12, line 111: looks queryparams logic is removed in the final method. believe we should include it

resources/sdk/purecloudgo/templates/configuration.mustache
```diff
@@ -506,16 +514,17 @@ func (c *Configuration) AuthorizeClientCredentials(clientID string, clientSecret
 func (c *Configuration) AuthorizeCodeGrant(clientID string, clientSecret string, authCode string, redirectUri string) (*AuthResponse, error) {
 	c.ClientID = clientID
 	c.ClientSecret = clientSecret
-	authHostRegex := regexp.MustCompile(`(?i)\/\/api\.`)
-	authHost := authHostRegex.ReplaceAllString(c.BasePath, "//login.")
```
HemanthDogiparthi12, line 510: here the way we are constructing the url , is different from gc cli. better we dont change this logic and use the same in getconfurl
for the basepath

resources/sdk/purecloudgo/templates/configuration.mustache
```diff
@@ -549,15 +558,15 @@ func (c *Configuration) AuthorizeCodeGrant(clientID string, clientSecret string,
 // RefreshAuthorizationCodeGrant requests a new access token for the authorization code grant.
 // The access and refresh tokens will be set automatically and API instances using this configuration object can continue to make authorized requests.
 func (c *Configuration) RefreshAuthorizationCodeGrant(clientID string, clientSecret string, refreshToken string) (*AuthResponse, error) {
-	authHostRegex := regexp.MustCompile(`(?i)\/\/api\.`)
```
HemanthDogiparthi12, line 552: same as top

HemanthDogiparthi12: and for the rest too

resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -106,20 +147,8 @@ func (c *APIClient) CallAPI(path string, method string,
...
-				urlString += fmt.Sprintf("%v=%v&", url.QueryEscape(strings.TrimSpace(k)), url.QueryEscape(strings.TrimSpace(v)))
-			}
-		}
-	}
-	urlString = urlString[:len(urlString)-1]
-	u, err := url.Parse(urlString)
-	if err != nil {
-		return nil, err
-	}
+	u := getConfUrl(c.configuration, pathName, "", c.configuration.Host)
```
HemanthDogiparthi12, line 151: what would be the value of  pathname here

the logic of urlString := path + "?" 

urlString = urlString[:len(urlString)-1]
	u, err := url.Parse(urlString)
	if err != nil {
		return nil, err
	}
	
	is also ignored in getconfurl method. this might create regression problems .

carnellj-genesys (approved):

resources/sdk/purecloudgo/templates/apiclient.mustache
```diff
@@ -96,6 +96,46 @@ func getPathValue(pc *ProxyConfiguration, key string) (string, bool) {
 	return "", false
 }
 
+func getConfUrl(c *Configuration, path string, extendedPath string, overrideHost string) *url.URL {
+	var uri *url.URL
+	gateWayConfiguration := c.GateWayConfiguration
+
+	uri, _ = url.Parse(fmt.Sprintf("https://api.%s%s", c.BasePath, extendedPath))
```
HemanthDogiparthi12, line 103: if we are not setting url explicitly previously, dont think we need to change that behaviour. same applies for login as well. as far as i see go sdk doesnt explcitly differentiate api and login url , since it has a single callapi method with "path" getting injected to it.  dont think we have to change that behaviour. just check for gateway params and override them if present.

resources/sdk/purecloudgo/templates/configuration.mustache
```diff
@@ -702,3 +715,66 @@ func (c *LoggingConfiguration) SetLogToConsole(logToConsole bool) {
 func (c *LoggingConfiguration) GetLogToConsole() bool {
 	return c.logToConsole
 }
+
+func (c *Configuration) GetGateWayConfiguration() string {
+	return getGateWayConfig(c.UserName)
+}
+
+func getGateWayConfig(profileName string) string {
```
HemanthDogiparthi12, line 723: there is no need of viper in go sdk.  this code is only for cli sdk client.

HemanthDogiparthi12 (approved):

PR #937, opened 2024-10-16
Ability to set API Host and Login Host (Platform API Javascript SDK)

themooks:
Platform API Javascript SDK:
Ability to set the API Host and the Login Host (Genesys Cloud regional hosts or custom gateways).

Introducing setCustomEnvironment method (Configuration and ApiClient classes) and login_host config file parameter.


themooks:
I will create a new code based on he Gateway configuration approach and else from CLI/Go

PR #940, opened 2024-10-25
[DEVTOOLING-665] Add path proxy config to Java SDK

themooks:
Add path proxy config to Java SDK


carnellj-genesys (approved):

themooks:
@dginty4 Following discussion in today's call, I will add test for this feature once we use an API Gateway or something more appropriate. serviceproxy.net is not an option (blocked by company) and considering that to pass a test we need to fail the request is not ideal test.

HemanthDogiparthi12 (approved):

PR #950, opened 2024-11-11
Add TS type for PureCloudRegionHosts

osvein:


dginty4 (approved):

resources/sdk/purecloudjavascript/templates/index.d.ts.mustache
```diff
@@ -2,6 +2,23 @@ import {{moduleName}} = require('{{projectName}}');
 
 declare module '{{projectName}}' {
 	export const ApiClient: ApiClientClass;
+	export const PureCloudRegionHosts {
```
HemanthDogiparthi12, line 5: we already have a export for the same in extensions folder. PureCloudRegionHosts.js. Use the same reference here too.

osvein: I'm not sure I understand. I'm aware that PureCloudRegionHosts.js contains the export. This PR doesn't add another export, it adds TypeScript types to the existing export. `declare module` is a TypeScript construct which adds type definitions to an existing JS module.

HemanthDogiparthi12 (approved):

PR #958, opened 2024-11-15
Add Http Client Inject

HemanthDogiparthi12:


package.json
themooks, line 1: I think that hpagent and https-proxy-agent don't need to be declared as dependencies here. They should be removed if it is the case.

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -204,6 +206,33 @@ class ApiClient {
...
 
+    setHttpClient(httpClient) {
+        if (!(httpClient instanceof HttpClient)) {
+            throw new Error("httpclient must be an instance of Abstract HttpClient. See DefualtHttpInstance for a prototype");
+        }
+        this.httpClient = httpClient;
+    }
+
+
+	setHttpAgentPaths(certPath, keyPath, caPath) {
```
themooks, line 217: The setHttpAgentPaths would need to have a switch/test on window === undefined, to have code for node (with require on fs) and code for browser.

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -1163,6 +1194,61 @@ class ApiClient {
...
+    	 */
+    	buildHttpClient() {
+    		if (this.httpClient) {
+    		    return this.httpClient;
+    		} else {
+    		    return this.defaultClient();
+    		}
+    	}
+
+        buildRequestOptions(url, method, headers, data, params) {
```
themooks, line 1209: Not critical. But the buildRequestOptions could be removed and its code moved to a RequestOptions constructor.

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -1163,6 +1194,61 @@ class ApiClient {
...
+                if (this.agentOptions) {
+                    proxyAgent = new https.Agent(this.agentOptions);
+                }
+
+                // If that.proxyAgent is available, use it (prioritize this over agentOptions)
+                if (this.proxyAgent) {
+                    proxyAgent = this.proxyAgent;
+                }
+
+                var httpClient =  new DefaultHttpClient();
```
themooks, line 1239: It could be good to avoid doing unnecessary new on DefaultHttpClient. I mean doing it once only.

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
```diff
@@ -1163,6 +1194,61 @@ class ApiClient {
...
+
+             request.headers =  headers;
+             return request;
+        }
+
+         defaultClient() {
+                // Initialize proxyAgent as null
+                let proxyAgent = null;
+
+                // If agentOptions are provided, create a new https.Agent with those options
```
themooks, line 1229: At this time, it appears that we can either set certificates or set a proxyUrl (the legacy use of proxy). It would be necessary to find the necessary setting/setup/code to have both certificates and proxy url (if it makes sense), and if not, document in README that it is one or the other.

resources/sdk/purecloudjavascript/templates/DefaultHttpClient.mustache
```diff
@@ -0,0 +1,51 @@
...
+
+// Default client is Axios
+class DefaultHttpClient extends HttpClient{
+
+    constructor() {
+        super();
+        this.proxy_agent;
+        this.timeout;
+        this._axiosInstance = axios.create({});
+        this._axiosInstance.interceptors.request.use((config) => {
```
themooks, line 13: Remove if not used at this stage.

resources/sdk/purecloudjavascript/templates/HttpRequestOptions.mustache
```diff
@@ -0,0 +1,53 @@
+export class RequestOptions {
```
themooks, line 1: Add headers.
Possibly add constructor to replace the buildRequestOptions method.

themooks (commented):
The new class and methods need to be defined in index.d.ts for typescript support/declaration.

Explanations need to be added to the README (at least how to set the certificates with the new methods, and ideally how to get both certificates and proxyUrl if it is doable).

HttpRequestOptions.mustache, DefaultHttpClient.mustache and abstractHttpClient.mustache will need an update on the openapi-generator to support these files (PR to merge). Or you could also rename these files to .js and put them in the sdk extensions subdirectory (where PureCloudRegionHosts.js is). That's personally my preference (as it doesn't need changes in openapi-generator)

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
HemanthDogiparthi12: done

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
HemanthDogiparthi12: done

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
HemanthDogiparthi12: updated the readme and addressed.

resources/sdk/purecloudjavascript/templates/DefaultHttpClient.mustache
HemanthDogiparthi12: done

resources/sdk/purecloudjavascript/templates/HttpRequestOptions.mustache
HemanthDogiparthi12: done

themooks (approved):

PR #980, opened 2025-01-08
Update Readme for MTLS

HemanthDogiparthi12:


themooks (commented):
"or if you using this in a browser"
You can remove this. https library is not supported in browser. setMTLSContents is also for nodejs only.

HemanthDogiparthi12:
> "or if you using this in a browser" You can remove this. https library is not supported in browser. setMTLSContents is also for nodejs only.

removed.

themooks (approved):

PR #986, opened 2025-01-15
Add an epilogue to all README files 

bbbco:
Add an epilogue to all README files detailing that contributions should not be made directly to the repo but on the commons repo


bbbco:
@HemanthDogiparthi12 this is what I was telling you about the other day

bbbco:
Bump!

PR #987, opened 2025-01-16
Updating asynchttpclient package

dginty4:


themooks (approved):
Looks fine to me. I think we also need to make the same update on purecloudjava-guest sdk, as it also gets published on maven with same cve alert.

dginty4:
> Looks fine to me. I think we also need to make the same update on purecloudjava-guest sdk, as it also gets published on maven with same cve alert.

Yes you're right, I'll update that too

PR #988, opened 2025-01-16
[NO-JIRA] .NetSDK - Create distinct Configuration/Configuration.ApiClient instances

themooks:
Configuration constructor currently leverages the ApiClient of the Default Configuration instance (static). The change is to facilitate the creation of Configuration and ApiClient instances that dot not rely on the Default Configuration singleton - allowing to maintain distinct ApiClients to different regions (or same region with different access tokens).

### DIFF
--- a/resources/sdk/pureclouddotnet/templates/Configuration.mustache
+++ b/resources/sdk/pureclouddotnet/templates/Configuration.mustache
@@ -40,6 +40,7 @@ namespace {{packageName}}.Client
         /// <param name="userAgent">HTTP user agent</param>
         /// <param name="configFilePath">Config file path</param>
         /// <param name="autoReloadConfig">AutoReloadConfig</param>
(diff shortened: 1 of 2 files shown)


carnellj-genesys (approved):
Should we be doing this in the other SDKs.

themooks:
> Should we be doing this in the other SDKs.

.Net SDK is the only one I know which had such behavior.
An "unusual" implementation of Configuration constructor and ApiClient constructor.
Creating a Configuration without ApiClient as input param lead to pointing to Default Configuration (static/singleton).
And creating an ApiClient without Configuration as input param lead to pointing to ApiClient of the Default Configuration.
Using the SDK, there is a way with additional code to finally get independent clients (what I described in the forum post on nuget problem) but it is lengthy.
The current change is just to make this process easier.

HemanthDogiparthi12 (approved):

PR #994, opened 2025-02-04
Feature/devtooling 1009 - Introduce PreHook and PostHooks JS SDK

HemanthDogiparthi12:


resources/sdk/purecloudjavascript/templates/index.d.ts.mustache
themooks, line 1: * setPreHook and setPostHook also need to be defined in DefaultHttpClient class.
* I think the hookFunction definition possibly needs a Promise - not totally sure but believe so: hookFunction: (config: any) => Promise<any>
* minor: could we have something like enableHooks instead of executeHooks

resources/sdk/purecloudjavascript/templates/README.mustache
themooks, line 1: * For any custom requirements like pre validations or post cleanups (for ex: OCSP and CRL validation), we can inject the prehook and posthook functions.
The SDK's default client will make sure the injected hook functions are executed.

* Async example?
async function PreHook(config) {
	try {
		console.log("Running Pre-Hook: Certificate Revocation Checks");
        // custom validation logics
		console.log("Certificate validated successfully.");
		return config;
	} catch (error) {
		console.error("Pre-Hook Validation Failed:", error.message);
		throw error; // Reject request if validation fails
	}
}

resources/sdk/purecloudjavascript/templates/ApiClient.mustache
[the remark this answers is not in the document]
HemanthDogiparthi12: moved the implementation specific logic to enableHooks and the call is made only during setposthook and setprehook functions. removed it from the constructor.

resources/sdk/purecloudjavascript/templates/index.d.ts.mustache
HemanthDogiparthi12: - Assume you mean the ts declaration. Updated it. As they are inherited from abstract, think they dont need to be defined again in the DefaultHttpClient class.
- Corrected the method signature.
- Modified the mehtod name to enableHooks.
- moved the implementation specific logic to enableHooks and the call is made only during setposthook and setprehook functions. removed it from the constructor.

themooks: Sorry, you were right on setPrehook, setPostHook in index.d.ts for DefaultHttpClient. I got confused with other abstract method. It is indeed not necessary. Only enableHooks was necessary

resources/sdk/purecloudjavascript/templates/README.mustache
themooks: After your change - for the Prehook example, it is not "return error;" but "throw error;"

resources/sdk/purecloudjavascript/extensions/DefaultHttpClient.js
themooks, line 1: I think the console.log in enableHooks can be removed.

resources/sdk/purecloudjavascript/templates/index.d.ts.mustache
themooks: (any | Promise<any> | void) can be at least simplified to (any | Promise<any>)

themooks (approved):

charliecon (approved):

PR #998, opened 2025-02-13
[Devtooling-1063] Logging when max retries hit

fergalmcl:


resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -889,6 +893,10 @@ public class ApiClient implements AutoCloseable {
                 if (e.getStatusCode() == 401 && shouldRefreshAccessToken) {
                     handleExpiredAccessToken();
                     return getAPIResponse(request, returnType, isAuthRequest);
+                } else if (e.getStatusCode() == 429) {
```
HemanthDogiparthi12, line 896: will this usecase ("e.getStatusCode() == 429") be covered in e.getStatusCode() >= 300 below ?

resources/sdk/purecloudjava/templates/ApiClient.mustache
```diff
@@ -829,6 +829,10 @@ public class ApiClient implements AutoCloseable {
             }
             return new ApiResponseWrapper<>(statusCode, reasonPhrase, headers, body, entity);
         }
+        else if (statusCode == 429) {
```
HemanthDogiparthi12, line 832: you can add String message = "error"; as a default message and then override it with "Max number of retries exceeded";

reduces code duplication

HemanthDogiparthi12 (commented):
I think you have to log the error when the sdk is sleeping until the retry-after is reached. and retrying again . 

look at this piece of code in apiclient.

private boolean waitBeforeRetry(long retryAfterMs){}

Adding a log here, solves the purpose of this ticket and gives more visibility to the user that SDK is in retry process. 

"message":"Rate limit encountered. Retry the request in [retryAfterMs] seconds"

HemanthDogiparthi12 (approved):

HemanthDogiparthi12 (commented):
Please do the same change for other SDKs as well.

PR #1031, opened 2025-04-07
Feature/devtooling 996 mtls

kavinbalagen:


resources/sdk/purecloudjava/extensions/connector/apache/ApacheHttpClientConnectorProvider.java
```diff
@@ -75,4 +90,46 @@ public ApiClientConnector create(ApiClientConnectorProperties properties) {
 
         return new ApacheHttpClientConnector(client, executorService);
     }
+
+    private static SSLContext createSSLContext(ApiClientConnectorProperties properties){
+        
+        String defaultTrustStore = "";
+        String keyStorePath = properties.getProperty(ApiClientConnectorProperty.KEYSTORE_PATH, String.class, null);
```
HemanthDogiparthi12, line 97: apart from setting a path , check if we can have a provision to set the entire keystore and truststore contents ( see mtlscontents in js sdk)

resources/sdk/purecloudjava/tests/ApiMTLSClientInjection.java
```diff
@@ -0,0 +1,109 @@
+package com.mypurecloud.sdk.v2;
```
HemanthDogiparthi12, line 1: modify the filename to have Test at the end

resources/sdk/purecloudjava/extensions/connector/apache/ApacheHttpClientConnectorProvider.java
kavinbalagen: hey hemanth, We can do this, But keystore and truststore contents are encrypted and unsure if copy paste will ideally copy and retain all the characters, also truststore by default is a specific location (cacerts file) inside Java installation folder that JVM will usually read. In our test example, we are overriding the cacerts file path, but ideally the user will store their mtls truststore inside the Java default cacaerts file which can include other installed truststore for the JVM. Hence copy-pasting might not be good approach here.

HemanthDogiparthi12 (approved):

PR #1360, opened 2026-07-08
[NO-JIRA] Identify models and operations involving discriminators dynamically

themooks:
Identify models and operations involving discriminators dynamically and quarantine or override

### DIFF
--- a/modules/builder/builder.ts
+++ b/modules/builder/builder.ts
@@ -35,7 +35,7 @@ const quarantineOperationIds: string[] = ['postGroupImages', 'postUserImages', '
 const quarantineModels: string[] = [];
 // Keep, Quarantine or Override Discriminator and Polymorphism (possible values: keep, quarantine, override)
 const defaultDiscriminatorManagement: string = 'quarantine';
-const keepDiscriminatorModels: string[] = ['ListValues']
+const keepDiscriminatorModels: string[] = ['ListValues'];
 // Override OperationId due to name conflict ("operationId", "x-purecloud-method-name")
 const overrideOperationIds: any = {};
 const aliasOperationIds: any = {
@@ -971,7 +971,7 @@ function manageDiscriminator(discriminatorManagement: string, keepDiscriminatorM
 			// Manage Discriminator
 			if (discriminatorManagement === 'override') {
 				// Override
-				// Remove discrimnator models and their children from Swagger
+				// Remove discriminator models and their children from Swagger
 				for (let modelName of modelsWithDiscriminator) {
 					if (swaggerDiff.newSwagger.definitions[modelName]) {
 						delete swaggerDiff.newSwagger.definitions[modelName];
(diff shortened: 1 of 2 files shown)


modules/builder/builder.ts
```diff
@@ -30,10 +30,12 @@ let _this: Builder;
...
+const quarantineOperationIds: string[] = ['postGroupImages', 'postUserImages', 'postLocationImages'];
+const quarantineModels: string[] = [];
+// Keep, Quarantine or Override Discriminator and Polymorphism (possible values: keep, quarantine, override)
+const defaultDiscriminatorManagement: string = 'quarantine';
+const keepDiscriminatorModels: string[] = ['ListValues']
```
sureshperiyappan: any specific reason for missing the ";" here ?

modules/builder/builder.ts
```diff
@@ -915,6 +933,164 @@ function forceCSVCollectionFormat(forceCSVCollectionFormatInTags: string[]) {
...
+				}
+			}
+		}
+		log.info(`Found Discriminator based Models: ${modelsWithDiscriminator.toString()}`);
+		log.info(`Found Discriminator Child Models: ${childDiscriminatorModels.toString()}`);
+		if (modelsWithDiscriminator.length > 0) {
+			// Manage Discriminator
+			if (discriminatorManagement === 'override') {
+				// Override
+				// Remove discrimnator models and their children from Swagger
```
sureshperiyappan: Should be "discriminator".

modules/builder/builder.ts
```diff
@@ -915,6 +933,164 @@ function forceCSVCollectionFormat(forceCSVCollectionFormatInTags: string[]) {
...
+					pathsAsString = pathsAsString.replace(regexConvertRef, '"#/definitions/JsonNode"');
+				}
+				swaggerDiff.newSwagger.definitions = JSON.parse(definitionsAsString);
+				swaggerDiff.newSwagger.paths = JSON.parse(pathsAsString);
+			} else if (discriminatorManagement === 'quarantine') {
+				// Quarantine
+				// Find models with a direct or indirect reference on modelsWithDiscriminator or childDiscriminatorModels
+				// Init with discriminator based models and their children
+				let modelsToQuarantine: string[] = [];
+				for (let modelName of modelsWithDiscriminator) {
```
sureshperiyappan: This is just:

let modelsToQuarantine = [...modelsWithDiscriminator, ...childDiscriminatorModels];

Same pattern repeats for searchModels.

any specific reason to use for?

modules/builder/builder.ts
```diff
@@ -915,6 +933,164 @@ function forceCSVCollectionFormat(forceCSVCollectionFormatInTags: string[]) {
...
+						delete swaggerDiff.newSwagger.definitions[modelName];
+					}
+				}
+				// Override references to Discriminator Models with JsonNode (generic object)
+				if (!swaggerDiff.newSwagger.definitions['JsonNode']) {
+					swaggerDiff.newSwagger.definitions['JsonNode'] = { type: ItemsType.Object };
+				}
+				let definitionsAsString = JSON.stringify(swaggerDiff.newSwagger.definitions);
+				let pathsAsString = JSON.stringify(swaggerDiff.newSwagger.paths);
+				for (let modelName of modelsWithDiscriminator) {
```
sureshperiyappan: In the override path, we delete childDiscriminatorModels from definitions but only replace $refs for the parent models. If anything else in the spec still points to a child (e.g. #/definitions/EmailNotification), that ref is now dangling and will break codegen.

I know override isn't the default mode so it won't blow up today, but it's a quick fix — just include childDiscriminatorModels in the replacement loop alongside modelsWithDiscriminator. Otherwise whoever tries override mode first is going to have a bad time debugging it.

modules/builder/builder.ts
themooks: No. A miss. But it is not mandatory to have it. I will add it.

modules/builder/builder.ts
themooks: ok

modules/builder/builder.ts
themooks: No specific reason. I just avoid spread in general, with typescript. Not in this case, but just as a "habit" when filling/setting an interface as it causes issues

sureshperiyappan (approved):

